PR #360, opened 2014-10-22
Exposed promisifyAll's defaultFilter through a property on the function.

marneborn:
Rather than replicating the promisifyAll defaultFilter in my custom filter, I'd like to call it.
This way I get any changes that are made automagically.
I've simply added a property on the function called defaultFilter pointing to the function.
promisifiableMethods still calls the function declared in the closure, so changing the property doesn't change the default behavior, only usage outside.

```javascript
Promise.promisifyAll(..., {
    filter : function (name, func, target) {
         if ( ! Promise.promisifyAll.defaultFilter(name, func, target) )
             return false;
        // my code
    }
});
```

I've added a section to the API.md describing this.
I didn't create any tests (there weren't any for filter yet...)
Let me know if you'd like me to add some.

There are big differences js/browser/bluebird.js, but I didn't do anything more than add one line to promisify.js and touch the API.md, then ran build and test, so I don't think that this is anything that I did.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1657,6 +1657,23 @@ function defaultFilter(name, func) {
 
 (`util` is bluebird util, not node.js util)
 
+The default filter is exposed as a property on the Promise.promisifyAll function.
+
+This is not what is run internally, so while you can change it, any changes won't affect the default behaviour
+
+To run the defaultFilter in your custom filter do:
+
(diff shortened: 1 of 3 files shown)


benjamingr:
@marneborn ping.

PR #423, opened 2015-01-01
bump deps, add missing uglify-js and modify browserify grunt task

heavyk:
to work with the newer version.


heavyk:
crap. I didn't notice that it updated testers too. I will fix

heavyk:
after fiddling with it a bit, if I run the tests manually `node --allow-natives-syntax ../mocharun.js mocha/when_spread.js` for example, it says all good, but not in the runner.. not sure what is going on. just close the request or whatever.. lolz shit always happens on upgrades. I didn't notice cause the lib works fine... awesome lib. very through. cheers

petkaantonov:
I am in the process of completely redoing the build system at the moment

PR #436, opened 2015-01-13
Promise.method : 0 arguments.length should remain 0

Joris-van-der-Wel:
I just hit this issue in my project. If you call a Promise.method function without arguments, its arguments.length is 1.


benjamingr:
Looks good to me.

While I'm looking at the code @petkaantonov  is there any reason promise syntax isn't used with mocha in these tests?

petkaantonov:
merged manually

PR #449, opened 2015-01-16
Fix captured trace length issue when decycling

spion:
Motivating example

If a chain of traces is made that looks like this

    c -> b -> a

then you add `d` and `e` to form two chains

    d -> c -> b -> a,
    e -> c -> b -> a

decycling from d will not update the length of e (and lead to a crash when `e.decycle()` is called later)

This patch fixes that by not relying on that length for anything else
except as a heuristic whether to decycle or not


spion:
todo: need to add a test for this

PR #454, opened 2015-01-19
fix Promise.map execute order

magicdawn:
demo.js
```js
global.Promise = require('bluebird');

function echoAsync (i) {
  var d = Promise.defer();
  setTimeout(function() {
    console.log("echoing %d",i);
    d.resolve();
  },1000);
  return d.promise;
}

Promise.map([1,2,3,4,5,6,7,8,9,10],function(i) {
  return echoAsync(i);
},{
  concurrency: 1
});
```

node demo.js results
```
echoing 1
echoing 10
echoing 9
echoing 8
echoing 7
echoing 6
echoing 5
echoing 4
echoing 3
echoing 2
```
this seems weird. I think we all prefer
```
echoing 1
echoing 2
echoing 3
echoing 4
echoing 5
echoing 6
echoing 7
echoing 8
echoing 9
echoing 10
```

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -102,7 +102,7 @@ MappingPromiseArray.prototype._drainQueue = function () {
     var values = this._values;
     while (queue.length > 0 && this._inFlight < limit) {
         if (this._isResolved()) return;
-        var index = queue.pop();
+        var index = queue.shift();
         this._promiseFulfilled(values[index], index);
     }
 };
--- a/test/mocha/when_map.js
+++ b/test/mocha/when_map.js
@@ -409,7 +409,7 @@ describe("when.map-test with concurrency", function () {
             });
         }, {concurrency: 5}).then(function() {
             assert(wasEmpty);
-            assert.deepEqual(b, [0, 1, 2, 3, 4, 10, 9, 8, 7, 6, 5]);
+            assert.deepEqual(b, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);
             done();
         });
 


petkaantonov:
The order is undefined as stated in the documentation, so that slow operations that happen to be at the start of the array are not bottlenecking faster operations that happen to be at the end.

Since you are actually iterating the array for side effects instead of mapping it you are looking for
`.each`/`Promise.each`:

```js
Promise.each([1,2,3,4,5,6,7,8,9,10],function(i) {
  return echoAsync(i);
});
```

petkaantonov:
Also you probably want to implement `echoAsync` as

```js
function echoAsync (i) {
     return Promise.delay(1000).then(function() {
          console.log("echoing %d",i);
     });
}
```

as `Promise.defer` is deprecated and leads to brittle and far more complex code than necessary

magicdawn:
I set the concurrency to 1 just for simulating...

Promise.each is sequenced, that means no concurrency.

the order is undefined, that's not right. Current the biger indexed item always `starts` earlier than smaller indexed, except for the first batch.

petkaantonov:
> I set the concurrency to 1 just for simulating...
> 
> Promise.each is sequenced, that means no concurrency.

So which one is it that you want - order or concurrency? They are mutually exclusive.

> the order is undefined, that's not right. Current the biger indexed item always starts earlier than smaller indexed, except for the first batch.

My point is that it doesn't matter because the order is undefined, code that relies on any particular order is incorrect. If the order was as you preferred, you'd (and many others) would surely write code that depended on the order and then wonder why everything breaks catastrophically as they add asynchronous operations or change the concurrency.

magicdawn:
> So which one is it that you want - order or concurrency? They are mutually exclusive.

order & concurency 

There may be some misunderstandings on `order`. I mean order in macro.

promise-map-order

benjamingr:
@magicdawn Petka understands what you're asking for - what he doesn't understand is why you want to change it.

`Promise.map` does not make any guarantees about iteration order - the way it iterates a sequence is not specified and the library might decide to iterate even items before odd ones tomorrow if there is a performance gain in that. The order `Promise.map` iterates the collection may change tomorrow in a minor version since the library makes no guarantee about it.

If you want a `Promise.map` with an order guarantee changing `Promise.map` here won't do you any good since it might change again tomorrow, next week or in two months - instead you're welcome to create your own method for mapping in that order with different guarantees and it won't break. Do feel encouraged to put it on npm.

petkaantonov:
It cannot possibly matter whether item 6 or item 20 is executed next, because any code you have in the map callback is broken if it relies on that.

magicdawn:
@petkaantonov  @benjamingr Got it.

PR #467, opened 2015-01-26
Fix for scope-related issue with node-webkit

matiaszanolli:
There's a call to the function setImmediate() on schedule.js that breaks compatibility with the Node-Webkit runtime.

It seems that Node-Webkit combines part of node's global scope with the one used by Chromium. Because of this, some functions have to be called as attributes of the global variable.

I made the replacement, this should not break compatibility with Node but it would be good to give it a try just in case.


petkaantonov:
Merged manually, thank you. It will be in the 2.9.3 release

Issue #487, opened 2015-02-08
Request for clarification of change made in 2.9.7


jimlloyd:
Hello. I have code using promisify that has worked without issues since at least version 2.3.11, through 2.9.6. With 2.9.7 the code fails.

I want to be able pass `require('bluebird').promisify` to a library that will create promisified methods of standard node async methods (unbound functions that depend on *this*). FYI, the library in question is [node-java](https://github.com/joeferner/node-java), with the feature [documented here](https://github.com/joeferner/node-java#promises). The feature was written to be workable with multiple Promises/A+ libraries. See the source file [testHelpers.js](https://github.com/joeferner/node-java/blob/master/testHelpers.js) to see (in comments) the various libraries that were tested. Most of the libraries (including bluebird) allowed me to pass a single promisify function and get valid results. Q was an exception, since the closest candidate function (denodify/nfbind) could only promisify a bound function, so I had to write a wrapper function promisifyQ show in testHelpers.js. It looks like that may now be necessary with bluebird.

Since this behavior change is recent, I want to ask for clarification. I see that Issue #470 may be the reason for the change in behavior, though that was specifically about promisifyAll, not promisify.

petkaantonov:
The change should not affect any existing code at all.

jimlloyd:
The change did affect my existing code, but my code arguably misused the API. The API documentation for Promise.promisify() clearly states:

*Note that if the node function is a method of some object, you need to pass the object as the second argument ...*

Despite this statement, I was able to pass unbound methods to promisify and get a valid promisified method. So, one way to look at this is that there has been a long-outstanding bug in which the implementation did not conform to the spec, and 2.9.7 fixes the bug. Another way to look at it is that the documentation was overly restrictive, and 2.9.7 inadvertently introduced an unnecessary change in a useful feature. Which way do you see it?

petkaantonov:
If you don't pass a second argument, the promisified method uses whatever `this` is to call the original function. There is no change to that in 2.9.7 and that still works the same.

jimlloyd:
Ok, it seems that I need to try to create a small test that reproduces the error I am now seeing. I see that your unit test 'when calling promisified function it should >  should use this if no receiver was given' that seems to be a good test for this, but there are some complexities to my particular case, including the fact that the promisification function is passed down into a node addon, so the actual call to promisify a method occurs in C++ code (lines 71-81 of https://github.com/joeferner/node-java/blob/master/src/javaObject.cpp).

[...]

PR #490, opened 2015-02-10
Clarify that promisify may be used on unbound methods.

jimlloyd:
@petkaantonov Per issue #487, you might want to update API.md along these lines.


petkaantonov:
merged manually

PR #495, opened 2015-02-12
Add neo-async to benchmark

suguru03:
I have developed  [Neo-Async](https://github.com/suguru03/neo-async).
I would appreciate if you could add it to benchmark list.


petkaantonov:
I don't want to endorse this library as it currently is as it appears to make blatantly false claims.

> I rewrote it and removed codes like setImmediate to boost the performance, so the detail logic may differ from Async.

That's not a valid optimization because it obviously changes observable functionality. So it's not a drop-in replacement for async. What you may think is detail logic is [actually an important feature for many](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony).

suguru03:
I apologize for my wrong claim and the delay.
I fixed source code and closed [this issue](https://github.com/suguru03/neo-async/issues/6).
Could you please check again this PR?

petkaantonov:
Nice, thank you for that. Landed in https://github.com/petkaantonov/bluebird/commit/3d0a2cbe37809e338d24f433839bd900e1d8f86e

suguru03:
Thank you so much.

PR #505, opened 2015-02-20
Adding caching for this._isFulfilled(). On an app I tested it in, the av...

marchant:
...erage time for Promise._settlePromiseAt() was 12ms, with that change, going down to 7ms, and total, going down from 900ms to 760ms, so about 18% performance increase

### DIFF
--- a/src/promise.js
+++ b/src/promise.js
@@ -565,6 +565,7 @@ Promise.prototype._reject = function (reason, carriedStackTrace) {
 Promise.prototype._settlePromiseAt = function (index) {
     var promise = this._promiseAt(index);
     var isPromise = promise instanceof Promise;
+    var isFulfilled = this._isFulfilled();
 
     ASSERT(async._isTickUsed);
     if (isPromise && promise._isMigrated()) {
@@ -572,11 +573,11 @@ Promise.prototype._settlePromiseAt = function (index) {
         return async.invoke(this._settlePromiseAt, this, index);
     }
     ASSERT(!this._isFollowing());
-    var handler = this._isFulfilled()
+    var handler = isFulfilled
         ? this._fulfillmentHandlerAt(index)
         : this._rejectionHandlerAt(index);
 
-    ASSERT(this._isFulfilled() || this._isRejected());
+    ASSERT(isFulfilled || this._isRejected());
 
     var carriedStackTrace =
         this._isCarryingStackTrace() ? this._getCarriedStackTrace() : undefined;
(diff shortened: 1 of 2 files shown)


marchant:
All test passed with node tools/test

petkaantonov:
First about the scheduler, MessageChannel is unacceptable because it is bugged in some browsers. If this is ok for your application then you can use `Promise.setScheduler` but it cannot be used in the general case.

I cannot see any difference from the isFulfilled caching alone, testing in node 0.12

marchant:
Promise.setScheduler, excellent, works like a charm, well done. I’m focusing tests on client, and especially mobile, it matters there. If there’s no difference on node than sounds like a win to me.

petkaantonov:
I don't have a mobile to test but if we are gonna do this we can do much better. We can load the bitfield once and use a macros like `IS_FULFILLED(bitField)` which compiles to e.g. `((bitField & 21319204) !== 0) so that testing all flags is fast and don't require a method call and a memory load.

marchant:
That sounds great!

benjamingr:
@marchant you should develop against 3.0 (the coming release) and not against master fwiw.

marchant:
Ok, thanks, I'll have to figure out how to do the browser build then!

petkaantonov:
`node tools/build --browser`

marchant:
Thanks, appreciated!

marchant:
What is 3.0 bringing?

petkaantonov:
@marchant 

- Website with better documentation, guides, tutorials, how-tos etc..
- New cancellation and all the issues which are labeled 3.0 in github
- A minor backward incompatible with Promise.delay, which puts the optional value argument as last instead of first
- Fair perf increase, especially for the parallel benchmark which now in 3.0 actually performs faster than the sequential one!

benjamingr:
@marchant also:

 - Better debugging experience with warnings that let you know about many common pitfalls (like a lint)
 - Cleaning the API.
 - Promises got a lot lighter (allocating them now is dirty cheap, cheaper than allocating an empty array)

petkaantonov:
Outdated example of warnings: http://imgur.com/a/t3xng (constructor return is removed now and several new ones are added that are not in the album)

petkaantonov:
> Promises got a lot lighter (allocating them now is dirty cheap, cheaper than allocating an empty array)

More accurately, the price in field count is same as allocating 2 empty arrays

PR #514, opened 2015-02-27
How about we kill the sync build from the docs?

benjamingr:
What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,7 @@
 
 # Introduction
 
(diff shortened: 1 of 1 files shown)


spion:
:+1:

petkaantonov:
The sync build doesn't exist in 3.0 and the docs make no reference it there.

benjamingr:
@petkaantonov but these are the docs I checked out the 3.0 branch o_0?

petkaantonov:
they will be deleted once website is ready

YuriSolovyov:
There will be website ? O__o

PR #515, opened 2015-02-27
added streamline benchmarks

bjouhier:


petkaantonov:
There should only be one file representing streamline per benchmark, and it should run on all platform as is without native modules. It looks like there is 6 files all implementing doxbee sequential.

I would also require an implementation of the parallel benchmark because it's very different kind of work-flow - to get a better overall picture.

bjouhier:
Streamline is a preprocessor. It can compile to 3 different runtimes (callbacks, generators and fibers). There is also a --fast option which produces faster code for deep call stacks (which does not shine at all in this bench because the stack is completely flat). This explains why I posted 6 files (the original source + 5 preprocessed files). The callbacks and generators modes don't require any native module but the fibers mode does.

So I propose to remove the fibers files and the generators-fast but I'd like to leave the streamline._js source. It won't hurt because the extension is not loaded by the bench.

I'll create the parallel variants and resubmit a PR.

petkaantonov:
So that leaves callbacks and generators, am I correct in assuming that the callbacks are faster but with unreadable output while generators are slower but the output is readable? Which one is typically used by users of streamline?

petkaantonov:
And yea we are missing a benchmark that represents deeper stacks, doxbee goes only 3 levels at best or so. If you have a production/realistic-looking-like routine (similar to the doxbee sequential benchmark) that you can share and has much deeper stack, we could create a third kind of benchmark with a shim version of that.

bjouhier:
I don't have stats but it looks like most people are using either callbacks or fibers. Callbacks are the fastest on this bench but they are outperformed by fibers when call stacks are deep. We bench our app in different modes (it is a simple config parameter). Our product is currently released in fibers-fast mode.

I've not heard of people using the generators modes. Probably because it does not run with node 0.10.

Streamline is also used in the browser. For example, vim was ported to the browser by transpiling with emscripten and streamline (callbacks mode).

I don't have a public realistic bench for deep calls. The benches that we use exercise our whole ERP product so I would need to recreate something new on the side.

I'm thinking of introducing a bluebird+generators mode. Should not be hard to implement and it could be a winner for us.

bjouhier:
PR should be ok now. Travis build failed but the error seems to be on the tests.

Streamline does not do as well on the parallel bench :disappointed:

petkaantonov:
Landed in https://github.com/petkaantonov/bluebird/commit/fd7da059c4566572f4e30db3c5c695bc5c2b8662

When 3.0 is released, and the website, the results will be published on the website. This will happen when I am not ridiculously busy with io.js :D sorry

bjouhier:
Thanks. I like it when you are _ridiculously busy with io.js_.

PR #516, opened 2015-03-01
3.0

marchant:
One bug fix and performance improvements


petkaantonov:
Thanks.

I need to say that you should not trust a profiler so blindly. I will merge it because it's a shorter way of saying the same thing, so it's more readable. But make no mistake, there is no way it actually improved any performance.

petkaantonov:
Landed in https://github.com/petkaantonov/bluebird/commit/84229705a69791af5c3af4354b68a0215e421e79 and  https://github.com/petkaantonov/bluebird/commit/9112e573111a1aa462147f146859b72d081738e9

PR #524, opened 2015-03-10
Added legitimate ASCII line endings to the regexp

Artazor:
Should it be there?


Artazor:
It prevented builds if source contains `'\r'` (also was false positives for `'\n'`)

petkaantonov:
The lines are split by \n so they cannot contain \n. \r is disallowed, configure your text editor and/or git correctly https://github.com/petkaantonov/bluebird/blob/master/.editorconfig

Artazor:
Sorry! -)

PR #546, opened 2015-03-25
Alias exec to nodeify

tgriesser:
I think nodeify is the only method in Bluebird I have issue with, is feels non-obvious compared to everything else - moreso now that there's probably a camp of people that might insist it be renamed `ioify` :P

But in all seriousness, it feels sort of clunky when a promise is exposed through library apis... 

```js
userModel.where('id', 1).fetchAll({withRelated: ['accounts']}).nodeify(cb)
```

"What's a nodeify? Why am I nodeifying a model?"

vs:

```js
userModel.where('id', 1).fetchAll({withRelated: ['accounts']}).exec(cb)
```

"Oh, execute the `fetchAll` with a callback"

I personally don't use callbacks here, but for those that do I think this offers a better UX - I believe mongoose uses `exec` as the method name for their callback handler which is where I originally got the naming.

### DIFF
--- a/src/nodeify.js
+++ b/src/nodeify.js
@@ -43,6 +43,7 @@ function errorAdapter(reason, nodeback) {
     }
 }
 
+Promise.prototype.asCallback = 
 Promise.prototype.nodeify = function (nodeback, options) {
     if (typeof nodeback == "function") {
         var adapter = successAdapter;


benjamingr:
Wouldn't it make more sense to call it `.toCallback`? It's a much more obvious name.

phpnode:
or even just `.callback()`?

tgriesser:
Yeah possibly `callback` - I guess it just felt clunky to have a method name describing the input vs what it's doing with the input. In this case it's executing the callback.

petkaantonov:
Just pure `exec` is probably not good, what about `asCallback`, `execCallback` etc? Damn naming is hard

benjamingr:
asCallback is good

tgriesser:
Yeah if not exec, I'd vote `asCallback` over `toCallback`

tgriesser:
Updated.

benjamingr:
Petka is probably going to ask you to merge into 3.0 :)

petkaantonov:
@benjamingr yes 2.x and 3.x are diverged so it would need to be duplicated in 3.0

tgriesser:
Any chance it could go into 2.x and then get cherry picked? What's the timeline on 3.0 by the way?

petkaantonov:
Yes this can go in 2.x. 3.x would have been done by now if I wasn't so busy with io.js :)

petkaantonov:
Can you also add the documentation to API.md in this same PR?

petkaantonov:
Landed in https://github.com/petkaantonov/bluebird/commit/b7601b33a37ca7fea5d22a64d2f6c9b203c391de released in 2.9.15

Issue #556, opened 2015-04-01
Promise.promisify() throws a type error under phantomjs


noahadams:
When calling `Promise.promisify()` under phantomjs (v 1.9.8 as packaged along with mocha-phantomjs v 3.5.3), a type error is thrown while copying descriptors:

```
     TypeError: Property description must be an object.
      at copyDescriptors (file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/bluebird/js/browser/bluebird.js:4706)
      at file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/bluebird/js/browser/bluebird.js:3400
      at file:///Users/nba/src/bluebird-promisify-phantomjs-bug/promisify-test.js:8
      at file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/mocha/mocha.js:4514
      at file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/mocha/mocha.js:4946
      at file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/mocha/mocha.js:5051
      at next (file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/mocha/mocha.js:4871)
      at file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/mocha/mocha.js:4881
      at next (file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/mocha/mocha.js:4816)
      at file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/mocha/mocha.js:4848
      at timeslice (file:///Users/nba/src/bluebird-promisify-phantomjs-bug/node_modules/mocha/mocha.js:6391)
```

phantomjs appears to supply the function with a prototypally inherited `callee` property which is not filtered out: https://github.com/petkaantonov/bluebird/blob/master/src/promisify.js#L15 but then fails to be copied.

This has only been a problem in phantom so far, but we're using it pretty extensively for CI purposes.

I've got a breaking test case here:
https://github.com/noahadams/bluebird-mocha-phantomjs-bug

~~and a branch ready to PR here:~~
~~https://github.com/noahadams/bluebird/tree/callee-property-filtering~~

PR: #557

PR #557, opened 2015-04-01
filter out `callee` for property copying in calls to `promisify()`

noahadams:
For compatibility with phantomjs as reported here: https://github.com/petkaantonov/bluebird/issues/556


petkaantonov:
Landed in https://github.com/petkaantonov/bluebird/commit/8a83e70730a72926daf967549a321f104c4c187f

noahadams:
Thanks!

Issue #562, opened 2015-04-04
Cancelation of promises chained from resolved promises


Artazor:
An inconsistency introduced by a48d63ff9c112c1d607dcd3acd86f170ed0a826c is caught by the following code:

```JavaScript

    specify("shoud be consistent for mid-chain breaking", function () {
        var called = 0;

        function test(p) {
            var chain;
            return chain = p.then(function() {
                // `p` has been fulfilled already at this point
                // and the nature of this fulfillment (immediate or eventual)
                // should not affect the future of the control flow
                chain.cancel();
            }).then(function(){
                called++;
            });
        }

        var immediate = Promise.resolve();
        var eventual = Promise.delay(0);

        test(immediate);
        test(eventual);

        return awaitLateQueue (function(){
            assert(called % 2 == 0);
            // either both should be called or both should not be called
        });
    });

```

Artazor:
*UPD*: I've updated the description of the issue

bergus:
I would make the test more strict and require `called == 0`. After you called `cancel()` on a chain, we expect that **no** more handlers in the chain are executed (`then`/`catch` handlers I mean, `oncancel`, `finally` etc obviously should be).

Artazor:
If this behavior is "by design", then it would be difficult to write tests.
For example, real requests could not be mocked with Promise.resolve(MockedResponse)

Artazor:
@bergus absolutely agree

petkaantonov:
This commit was reverted already

PR #563, opened 2015-04-04
added failing test case for mid-chain cancellation

Artazor:
Failing test case for issue #562

### DIFF
--- a/test/mocha/cancel.js
+++ b/test/mocha/cancel.js
@@ -24,13 +24,15 @@ describe("Cancellation", function() {
         });
     });
 
-    specify("shoud be consistent for mid-chain breaking", function () {
+    specify("chain cancellation shoud be history-agnostic", function () {
         var called = 0;
 
         function test(p) {
             var chain;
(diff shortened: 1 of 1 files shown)


test/mocha/cancel.js
```diff
@@ -24,6 +24,31 @@ describe("Cancellation", function() {
         });
     });
 
+    specify("shoud be consistent for mid-chain breaking", function () {
```
bergus: I wouldn't call this "mid-chain". Yes, it cancels while the chain is currently executing, but that's basically what we have `cancel` for.
"mid-chain" means to me that you have a chain `b = a.then(…), c = b.then(…)` and `b` is cancelled.

Artazor:
Now all is ok. Should I perform any rebase magic to the repo?

test/mocha/cancel.js
Artazor: I've changed wording now it is "chain cancellation should be history-agnostic"

Artazor:
Become irrelevant

PR #575, opened 2015-04-13
Rewording the `concurrency` limit option doc

thefourtheye:
1. Rewording the `concurrency` limit option documentation

2. Updating the `concurrency` link in `filter` to point to the `concurrency` limit section in `map`, instead of `map` itself.


petkaantonov:
Thank you!

PR #577, opened 2015-04-13
Should swap checks

Artazor:
stackFramePattern can be null (so it will throw on `.test`) even if line === NO_STACK_TRACE.
Faced with this problem under phantomjs

### DIFF
--- a/src/debuggability.js
+++ b/src/debuggability.js
@@ -318,8 +318,8 @@ function cleanStack(stack) {
     var ret = [];
     for (var i = 0; i < stack.length; ++i) {
         var line = stack[i];
-        var isTraceLine = stackFramePattern.test(line) ||
-            NO_STACK_TRACE === line;
+        var isTraceLine = NO_STACK_TRACE === line ||
+            stackFramePattern.test(line);
         var isInternalFrame = isTraceLine && shouldIgnore(line);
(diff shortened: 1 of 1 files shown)


benjamingr:
A test for the failing case would be nice

Artazor:
@benjamingr I'm afraid that this is not reproducible in your testing environment. As far as I know Bluebird probes the stackTraceFormat by creating an Error and inspecting its format (selects the right pattern). Both console and live-browser tests

    $node ./tools/test.js
and
    $node ./tools/test.js --browser

do not show any errors (all test are OK)

However when I run the same tests under the PhntomJS with longStackTrace enabled - I'm getting some tests failed with error "unable to read property `test` on null" exactly at the line I've specified (in debuggability.js).

Inspecting the actual value of `line` variable shows that it is exactly `"     (No stack trace)"` (the value of the NO_STACK_TRACE constant) that means that the reorder of the '||' parts will solve the problem.

I could try to setup a minimal testing environment with the PhantomJS included, but it does not sound like a few minutes job.

petkaantonov:
Landed in https://github.com/petkaantonov/bluebird/commit/e04522feea5055fc979b3f5aba6df1dfab4f5346

PR #578, opened 2015-04-13
Added onRejected to Promise.prototype.spread()

### LINKED ISSUE #576: Promise.prototype.spread signature changed
According to [2.x docs](https://github.com/petkaantonov/bluebird/blob/master/API.md#spreadfunction-fulfilledhandler--function-rejectedhandler----promise) we should be able to pass both `onFulfilled` and `onRejected` handlers to `.spread()`. Now in 3.0 it [looks like](https://github.com/petkaantonov/bluebird/blob/3.0/src/promise.js#L115-L120) that only `onFulfill` is expected. Is it by design?

Artazor:
Fixes #576 

This is only a trial to restore an original documented behavior for `spread`.
Should add more tests.
If it's behavior has changed by design, please confirm that explicitly.
Or have I've missed anything?

### DIFF
--- a/src/promise.js
+++ b/src/promise.js
@@ -112,11 +112,11 @@ Promise.prototype.done = function (didFulfill, didReject) {
     promise._setIsFinal();
 };
 
-Promise.prototype.spread = function (fn) {
-    if (typeof fn !== "function") {
-        return apiRejection(FUNCTION_ERROR + util.classString(fn));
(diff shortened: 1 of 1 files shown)


Artazor:
Sorry, need to fix a constant FUNC_ERROR (accidentally replaced it with the value)

Artazor:
**UPD**: Recommitted as single commit

petkaantonov:
Yes, it's by design

PR #579, opened 2015-04-13
define index variable out of the loop

capaj:
just something a fellow redditor(http://www.reddit.com/user/Kollektiv) spotted


benjamingr:
Why would this make any difference? `var` defines a variable in the scope of the enclosing function.

petkaantonov:
@benjamingr `var` is treated as if it was block scoped (enforced by jshint). 

@capaj `index` is not a loop index but index into array, there appears to be no need to make it more "visible" than it is.

capaj:
:+1: thought it would be more performant. I was wrong.

PR #593, opened 2015-04-21
Fix error when running in nw.js

skeary:
Hi Petka,

I noticed an issue with running the latest version inside  nw.js (node-webkit). In schedule.js `process.versions.node` is assumed to exist. In node-webkit, however, it doesn't so the call to `process.versions.node.split(".")` fails.

Here's a pull request which I believe fixes the issue. In nw.js `process.versions["node-webkit"]` is used instead. As you can see from the nw.js changelog (https://github.com/nwjs/nw.js/blob/master/CHANGELOG.md), v0.9.0 of nw.js uses Node.js v0.11 so the new nwjs-related check compares against version 0.8 of nw.js.

Cheers,
Simon

### DIFF
--- a/src/schedule.js
+++ b/src/schedule.js
@@ -4,9 +4,15 @@ var noAsyncScheduler = function() {
     throw new Error(NO_ASYNC_SCHEDULER);
 };
 if (require("./util.js").isNode) {
-    var version = process.versions.node.split(".").map(Number);
-    schedule = (version[0] === 0 && version[1] > 10) || (version[0] > 0)
-        ? global.setImmediate : process.nextTick;
+    if (typeof process.versions["node-webkit"] !== "undefined") {
+        var version = process.versions["node-webkit"].split(".").map(Number);
+        schedule = (version[0] === 0 && version[1] > 8) || (version[0] > 0)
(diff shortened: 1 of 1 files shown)


skeary:
Sorry, yes, I don't think that it the best way to check for the existence of the version. I've modified the line to do what I think is a better check?

petkaantonov:
This is too verbose, perhaps 

```js
var version = process.versions.node || process.versions["node-webkit"];
version = version.split(".").map(Number);
```

petkaantonov:
I cannot find any reason to believe this is true in nw codebase, in fact https://github.com/nwjs/nw.js/blob/c636f94c0e0cbdfd52fc89d095414585f6a16275/src/resources/pages/nw_version.html

closing

skeary:
I've just checked again and yes, you're completely right, the fix isn't needed. I've realised what's going on... I had based the app of [requirejs-nodewebkit-template-project](https://github.com/majodev/requirejs-nodewebkit-template-project) and, if you look at [app.js](https://github.com/majodev/requirejs-nodewebkit-template-project/blob/master/app.js) in that project you can see it deletes process.versions.node.

Sorry!
Simon

PR #596, opened 2015-04-24
[docs] contradiction

timruffles:
this was confusing

### DIFF
--- a/API.md
+++ b/API.md
@@ -1796,7 +1796,7 @@ The resolver function is passed a callback that expects to be called back accord
 Using manual resolver:
 
 ```js
-// TODO use real library that doesn't expose prototypes for promisification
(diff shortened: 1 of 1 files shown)


benjamingr:
Thanks!

petkaantonov:
? this changes the meaning

benjamingr:
@petkaantonov yes, but why would you want to switch to a library that doesn't expose prototypes? 

The old meaning made no sense (at least from your PoV)

petkaantonov:
the todo is for the docs that are currently using fake example but should use a realistic one instead

benjamingr:
@petkaantonov it looked like a really opinionated TODO on what the user reading the docs should do and not a TODO for the docs themselves - as if you were suggesting replacing the library people were using with a bad one.

PR #607, opened 2015-05-01
DONT MERGE. Question re: promisifying request.js

philosoralphter:
I am wondering if this business about having to use the convenience methods with Async appended to the method name: `request.getAsync()` and `request.postAsync()` etc. rather than being able to use request(...).  Can anyone help out and lets update or expand on this section?

### DIFF
--- a/API.md
+++ b/API.md
@@ -1444,9 +1444,18 @@ Promise.promisifyAll(require("mongoose"));
 ```js
 // Request
 var Promise = require("bluebird");
+
+//Option 1: promisify the convenience methods
 Promise.promisifyAll(require("request"));
-// Use request.getAsync(...) not request(..), it will not return a promise
-//Is the above comment still accurate?
+    //This Promisifies the convenience methods (like `request.get()` and `request.post()`)
+    //It appends "Async" to the method names so use, eg: 
+        //`request.getAsync( {{request-options}} ).then(...)`. 
+        //Do not use request(..), it will not return a promise--for more info see PR #607 (https://github.com/petkaantonov/bluebird/pull/607)
+    
+//Option 2: Promisify the request() method
+Promise.promisify(require("request"));
+    //Now you can use `request( {{request-options}} ).then(...)`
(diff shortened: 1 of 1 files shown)


bendrucker:
Yes, it's correct and accurate. Docs PRs are always welcome. `request` is a function with methods like `get` on it. promisifyAll` adds the `*Async` methods, but you obviously can't overwrite the reference to `request`. 

Your other choice appears elsewhere in the docs as an example:

```js
var request = Promise.promisify(require('request'));
```

If you're so, send a PR adding that as an alternative to promisifying the methods and briefly explain the difference

philosoralphter:
Hey Ben.  Thanks for the reply. Not sure I know what the difference would be between promisify and promisifyAll.  I'll look into it and try to make this section clearer this week.  Thanks again.

bendrucker:
Ah, that explains it. Here's a super simple implementation of each:

```js
function promisify (fn, receiver) {
  return function promisified () {
    var args = [].slice.call(arguments)
    return new Promise(function (resolve, reject) {
      fn.call(receiver, args.concat(function (err, result) {
        if (err) return reject(err)
        resolve(result)
      }))
    })
  }
}

function promisifyAll (object) {
  for (var key in object) {
    if (typeof object[key] === 'function') {
      object[key + 'Async'] = promisify(object[key], object)
    }
  }
  return object
}
```

Bluebird's implementation is quite a bit more complicated in order to address all kinds of trickier cases. But this is the gist.

philosoralphter:
Ahhh, I see.  That makes sense now.  I'll edit my PR

philosoralphter:
Perhaps I should have made a new PR...Anyway, thanks for your help @bendrucker !

analog-nico:
@philosoralphter The request project has a few [tests](https://github.com/request/request/blob/master/tests/test-promise.js) that reflect what your are currently describing in your PR. FYI, check out [request/request-promise](https://github.com/request/request-promise), too.

PR #613, opened 2015-05-09
WIP 3.0docs

benjamingr:
WIP, working on the documentation pages, opening a PR so we don't conflict and for feedback.


petkaantonov:
I have merged this already

Btw a Title's Casing is determined like this:

```js
var toTitleCase = (function() {
    var noCapitalizeWords = "the a an for and nor but or so to in of with on from at as".split(" ");
    var noCapitalizeWordsMap = {};
    for (var i = 0; i < noCapitalizeWords.length; ++i) {
        noCapitalizeWordsMap[noCapitalizeWords[i]] = noCapitalizeWordsMap;
    }
    function capitalize(word, isFirstOrLast) {
        if (!isFirstOrLast && noCapitalizeWordsMap[word] === noCapitalizeWordsMap) {
            return word;
        }
        return word.charAt(0).toUpperCase() + word.slice(1);
    }
    return function toTitleCase(str) {
        var words = (str + "").trim().toLowerCase().split(/\s+/g);
        var last = words.length - 1;
        words[0] = capitalize(words[0], true);
        for (var i = 1; i < words.length - 1; ++i) {
            words[i] = capitalize(words[i], false);
        }
        words[last] = capitalize(words[last], true);
        return words.join(" ");
    }
})();
```

PR #617, opened 2015-05-12
Correct Markdown Syntax

Mindstormer619:
Small change to correct formatting of one line.


benjamingr:
Thanks, we'd appreciate help improving and debugging the new docs if you feel like it :)

They're at https://github.com/petkaantonov/bluebird/tree/3.0docs/docs/docs some parts are missing but most should be there :)

Mindstormer619:
You're welcome! I'm having a look at them as we speak. Just got started on Promises with node... looks promising :stuck_out_tongue_winking_eye:

PR #620, opened 2015-05-15
Add benchmark for Kefir (RxJS/Bacon competitor)

pjeby:


pjeby:
(FYI, the Travis build failure appears spurious, caused by a connection failure to Sauce Labs and a build timeout long before it could actually run any tests.  The non-Sauce builds worked fine.)

PR #622, opened 2015-05-18
Fix leaks from `tryCatchTarget`, `errorObj` and `NEXT_FILTER`

dohse:
In applications that make sporadic use of bluebird a callback might be retained in `tryCatchTarget`. This leads to all objects associated with that promise/callback chain being retained. Avoid this by resetting `tryCatchTarget` before calling it.

This is also the case for objects in `errorObj` and `NEXT_FILTER`. As errors might only occur occasionally this might retain objects associated with errors.

Retention in chrome heap profiler

### DIFF
--- a/src/catch_filter.js
+++ b/src/catch_filter.js
@@ -39,7 +39,8 @@ CatchFilter.prototype.doFilter = function (e) {
         if (itemIsErrorType && e instanceof item) {
             var ret = tryCatch(cb).call(boundTo, e);
             if (ret === errorObj) {
-                NEXT_FILTER.e = ret.e;
+                NEXT_FILTER.e = errorObj.e;
+                errorObj.e = null;
                 return NEXT_FILTER;
             }
             return ret;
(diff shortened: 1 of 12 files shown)


ruimarinho:
@dohse, can you share the methodology used for these findings?

petkaantonov:
setting `tryCatchTarget` to null makes sense but the others are not worth it

dohse:
@ruimarinho I added a patch containing a script and a screen shot to this PR. This should allow you to reproduce my findings.

dohse:
@petkaantonov `tryCatchTarget` is my main concern. I would consider fixing error objects as well though, as they are less often updated as `tryCatchTarget`.

benjamingr:
Thanks for the PR, why the dependency and `retention.js`? That's a pretty heavy dependency to introduce...

dohse:
@benjamingr the last commit is only intended for illustration, not for merge

dohse:
I am not sure whether `rentention.js` should be turned into a test case. I can be done with weak maps, `--expose-gc` and calls the `global.gc()`, but I am not sure whether that is reliable.

benjamingr:
@dohse at the very least it should be a dev dependency and probably an optional one.

dohse:
I want to remove the `retention.js` commit before this PR is merged.

dohse:
I removed the heapdump script here and put it into https://github.com/dohse/bluebird/tree/retention-script
@ruimarinho @petkaantonov @benjamingr what is you conclusion?

benjamingr:
@dohse can you squash?

dohse:
Done

petkaantonov:
https://github.com/petkaantonov/bluebird/commit/85e0d3423f676d43deac28e089338245f5f50130

PR #638, opened 2015-06-01
Warn when Promise.resolve is passed a function

benjamingr:
Warn when users pass `Promise.resolve` a callback returning function and expect it to return a promise. 

This was asked in Stack Overflow at least a dozen times (most recent http://stackoverflow.com/questions/30565855/can-bluebird-promise-work-with-redis-in-node-js). Apparently it's also an issue with new developers in code bases.

### DIFF
--- a/src/promise.js
+++ b/src/promise.js
@@ -174,6 +174,11 @@ Promise.cast = function (obj) {
         ret._setFulfilled();
         ret._rejectionHandler0 = obj;
     }
+    if(debug.warnings() && typeof obj === "function"){
+        ret._warn(".resolve() was passed a function and not a value. " +
+        "Note that this does not convert APIs to use promises. " + 
+        "See http://bluebirdjs.com/docs/api/promisification.html");
+    }
     return ret;
 };
 
--- a/src/promise.js
+++ b/src/promise.js
@@ -174,10 +174,12 @@ Promise.cast = function (obj) {
         ret._setFulfilled();
         ret._rejectionHandler0 = obj;
     }
-    if(debug.warnings() && typeof obj === "function"){
-        ret._warn(".resolve() was passed a function and not a value. " +
+    if(typeof obj === "function" && debug.warnings()) {
+        ret._warn(
+        ".resolve() was passed a function and not a value. " +
         "Note that this does not convert APIs to use promises. " + 
-        "See http://bluebirdjs.com/docs/api/promisification.html");
+        "See http://bluebirdjs.com/docs/api/promisification.html"
+        );
     }
     return ret;
 };


petkaantonov:
Do you have more links? I have literally never seen this and it's very normal and common and intended to pass values to `Promise.resolve`, it's not just for converting between implementations.

benjamingr:
@petkaantonov I can search if you'd like - it's perfectly fine to pass values to Promise.resolve but it makes no sense to pass a _function_ to `Promise.resolve` - this is what it checks.

The other issue except `Promise.resolve(fn)` is passing `Promise.resolve(fn())` which is just as bad but can't be tested.

bergus:
Well functions are values as well :-) Not necessarily an error if you're deep into functional programming.

But you're right, most of the cases where this happens are likely mistakes where people forgot an invocation, or wanted to use the `Promise` constructor or `Promise.try`.

benjamingr:
Right, this is a warning and not an error. So far I've never seen a user doing this and getting what they meant - if you have or are aware of code that does this in production - that of course changes everything.

src/promise.js
```diff
@@ -174,6 +174,11 @@ Promise.cast = function (obj) {
         ret._setFulfilled();
         ret._rejectionHandler0 = obj;
     }
+    if(debug.warnings() && typeof obj === "function"){
```
petkaantonov: Whitespace should be: `if () {`

src/promise.js
```diff
@@ -174,6 +174,11 @@ Promise.cast = function (obj) {
...
+        ret._warn(".resolve() was passed a function and not a value. " +
+        "Note that this does not convert APIs to use promises. " +
```
petkaantonov: The strings should all have the same indentation level e.g.:

```js
        "" +
        "" +
        "";
```

petkaantonov: Also I think the static warn method should be used for correct stack trace: `debug.warn(...)` instead of `ret._warn`, cannot check it atm.

phpnode:
Uh, this doesn't seem like it should trigger a warning, as @bergus says, functions are values. I've never seen anyone make this mistake either and I feel like the warning will introduce more confusion because a user will assume that they're doing something wrong when it's perfectly legitimate.

benjamingr:
>  I've never seen anyone make this mistake either and I feel like the warning will introduce more confusion because a user will assume that they're doing something wrong when it's perfectly legitimate.

I've seen dozens of cases where people used `resolve` incorrectly like this and I've never seen a case where you actually wanted to resolve a function.

phpnode:
@benjamingr the code which calls `Bluebird.resolve()` does not necessarily know or care about the type of the value being resolved, it could be a function or object or thenable or anything else. I don't think this is something to warn about, because it applies to *any* standard promise implementation and because the user already observes that they get the "wrong" value back. It's something that is better solved through big notices in the docs.

benjamingr:
@phpnode Can you show me anyone `resolve`ing with a function intentionally in the entirety of GH? That would go a long way to convince me this should not be a warning.

Bluebird is always about pragmatic help to the user over theoretical correctness. I've seen people do this by mistake plenty, showing someone who does this intentionally would help.

bergus:
@phpnode I second this warning. While functions are valid values, this is almost always a developer mistake.

src/promise.js
benjamingr: > Checking for function should be done first.

Why? I figured debug.warnings() would be inlined as false so that'd be slightly faster. Changing just curious.

spion:
I don't see the utility for this. But I do acknowledge it can be useful as training wheels for new users. I'm not sure if we really want that though

Here is a contrived example that would argue against it :P 

function makeEndpoint(resource) {
  var url = cache[resource];
  if (url) return params => get(url, params)
  else return getLocation(resource).then(url => params => get(url, params))
}

function fetch(resource, id) {
  return Promise.resolve(makeEndpoint(resource))
    .then(endpoint => endpoint({id: id}))
}

petkaantonov:
There is also no less than 3 different methods [in Q](https://github.com/kriskowal/q/wiki/API-Reference#promise-for-function-methods) to work with promises that are resolved with functions. This tells me it is possible there could be some common work flows where promises are resolved with functions.

benjamingr:
@petkaantonov this is because of the theoretical nature of Q and Mark Miller's work, not because of non-contrived use cases.

spion:
@benjamingr For observables I know of good examples, e.g. a (flux) store can be implemented as a [scan operation](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/scan.md) over an observable of state transforming functions with the initial state as the seed

benjamingr:
I'll make you a deal. At the moment I don't think anyone is using it, if people start using it and complaining about it - we can revert this. The example you linked to doesn't have a `Q()` (Promise.resolve) either.

I believe that we'll get nothing but happy users but I'm open to changing it if people actually use it this way.

spion:
Oh yeah, I agree that many beginners will be happier, but I don't know how to provide that without annoying advanced users :)

benjamingr:
I claim these "advanced users" are the same people who expect Promise.race
with `[]` to pend forever :D

This is (another) opinionated choice for the sake of 99.9% of users, if an
'advanced' user complains they can always turn the warning off or open an
issue here.

spion:
I'd estimate that number of users that will benefit to be 50% or less, but whatever.

Lets not get into `Promise.race` here though :) I can make a realistic example of why it might be a good idea...

petkaantonov:
> This is (another) opinionated choice for the sake of 99.9% of users, if an
'advanced' user complains they can always turn the warning off or open an
issue here.

There is no fine grained control over which warnings are enabled so "you can always turn it off" cannot be used as justification for a warning.

benjamingr:
I meant the same way people think "returning a value from the constructor
although a choice is their birth right" or "passing promise to `then` is
legit" sort of way.

phpnode:
> Can you show me anyone resolveing with a function intentionally in the entirety of GH? That would go a long way to convince me this should not be a warning

Fair play, I can't actually. After looking through some private client projects I found two occurrences where this pattern is used, both are part of query builders, where you can use a pattern like this:
```js
let result = await select("*").from("someTable").where({someField: conditional});
```
In this case `conditional` can be either another query, a function, a normal value or a promise which resolves to a normal value.

spion:
@phpnode I found such an example but @benjamingr probably missed it as I added by editing my message. Its in kraken-next:

https://github.com/totherik/kraken.next/blob/3c8b9633ef89770b466ebe52336c3e36420f37f3/lib/bootstrap.js#L21

benjamingr:
@spion that's all I needed to see. Thanks.

PR #649, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "bluebird",
-  "version": "2.9.27",
   "homepage": "https://github.com/petkaantonov/bluebird",
   "authors": [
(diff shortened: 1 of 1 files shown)


petkaantonov:
I don't see any reason to remove it. It is currently ignored by bower so there is no harm in having it, and it might even be used in the future.

Also, this file is automatically generated with a release script.

kkirsche:
Also just an FYI there are discussions to move bower to NPM anyway so I wouldn't really plan to support it for ever: bower/bower#1520

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

PR #658, opened 2015-06-15
Prevent Electron-based applications from being detected as node.js

paulcbetts:
If we are in an [Electron-based application](http://electron.atom.io), the browser process should have node.js Zen Nature, while other ones (renderer) should be treated as if we are in a browser

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -293,7 +293,8 @@ var ret = {
     hasDevTools: typeof chrome !== "undefined" && chrome &&
                  typeof chrome.loadTimes === "function",
     isNode: typeof process !== "undefined" &&
-        classString(process).toLowerCase() === "[object process]" && (!process.type || process.type === 'browser'),
+        classString(process).toLowerCase() === "[object process]" &&
+        (!process.type || process.type === "browser"),
     isNativeFunctionMethod: isNativeFunctionMethod
 };
 ret.isRecentNode = ret.isNode && (function() {
--- a/src/util.js
+++ b/src/util.js
@@ -293,7 +293,7 @@ var ret = {
     hasDevTools: typeof chrome !== "undefined" && chrome &&
                  typeof chrome.loadTimes === "function",
     isNode: typeof process !== "undefined" &&
-        classString(process).toLowerCase() === "[object process]",
+        classString(process).toLowerCase() === "[object process]" && (!process.type || process.type === 'browser'),
     isNativeFunctionMethod: isNativeFunctionMethod
 };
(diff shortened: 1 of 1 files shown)


benjamingr:
lgtm.

petkaantonov:
If it claims to be node it should work with the node apis that are being used. Why would it not work?

paulcbetts:
> If it claims to be node it should work with the node apis that are being used. Why would it not work?

I don't have the original crash trace, but it would crash on startup here: https://github.com/petkaantonov/bluebird/blob/31d02b46a750151ebba072a8ec455f818c1d0844/src/async.js#L19 with `schedule` being null.

This is probably because https://github.com/petkaantonov/bluebird/blob/634af0e27ff4faab62c6c5bfd105527abcf8b06e/src/schedule.js#L7 doesn't really grok things that are node but also have MutationObservers.

Philosophically, in the render process, we should try to think of it as Chrome as much as possible, this usually results in The Right Thing happening, rather than embracing node.js and trying to constantly patch away the "Oh, node.js! No DOM here!" type of decisions.

petkaantonov:
Its impossible for `schedule` to be null

paulcbetts:
> Its impossible for schedule to be null

¯\_(ツ)_/¯

Here's the minified stack trace, this crashes Electron 100% of the time for us.

```
TypeError: Cannot read property 'isStatic' of undefined
  at new Async (https://A URL/js/libs/bluebird-2.9.12.v1434403486.js:69:17)
  at Object.2../queue.js (https://A URL/js/libs/bluebird-2.9.12.v1434403486.js:157:18)
  at s (https://A URL/js/libs/bluebird-2.9.12.v1434403486.js:29:579)
  at https://A URL/js/libs/bluebird-2.9.12.v1434403486.js:29:630
  at module.exports (https://A URL/js/libs/bluebird-2.9.12.v1434403486.js:2119:13)
  at Object.4../promise.js (https://A URL/js/libs/bluebird-2.9.12.v1434403486.js:244:39)
```

petkaantonov:
The latest version is 2.9.30, not 2.9.12.

paulcbetts:
@petkaantonov :cool:

petkaantonov:
What does that mean? You are testing with a very old version while making a PR to master.

paulcbetts:
@petkaantonov We're just gonna maintain a patched version then, thanks for having a look!

benjamingr:
Paul, that makes no sense, what's preventing you from using 2.30?

PR #664, opened 2015-06-19
should preserve domain of the callback function

dimichgh:
currently, due to sharing of the promise, the caller that does the
resolve/reject passes its domain to other callbacks causing unexpected
behavior.

### DIFF
--- a/src/promise.js
+++ b/src/promise.js
@@ -217,19 +217,6 @@ Promise.prototype._then = function (
         if (!haveInternalData) ret._setIsMigrated();
     }
 
-    // preserve domain context if any
-    if (process.domain) {
-        if (didFulfill) {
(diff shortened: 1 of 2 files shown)


benjamingr:
Hehe, `Function#bind` in bluebird. How would this work with bluebird's [`.bind`](https://github.com/petkaantonov/bluebird/blob/master/API.md#binddynamic-thisarg---promise) for context preservation?

dimichgh:
It seems process.bind is smart enough to keep bound 'this' reference.
I have added tests for that.

petkaantonov:
Can you remove the implementation from the PR and just leave the tests?

A full fix will have to remove all the domain code from async.js and also it needs to be implemented more efficiently.

dimichgh:
sure, let me edit it out.

PR #671, opened 2015-06-24
Fix constant refresh causes IE 10 to leak memory

zomaish:


src/captured_trace.js
```diff
@@ -452,7 +452,12 @@ var fireGlobalEvent = (function() {
         } catch (e) {}
         if (!customEventWorks) {
             try {
-                var event = document.createEvent("CustomEvent");
+                var event;
+                if (document.createEventObject) {
```
davydog187: possibly this would be better split out as a utility function since its used twice? the comment would go nicely there.

benjamingr:
@zomaish any update?

zomaish:
I reached out to Microsoft for an explanation and waiting for a response. I will reach out again and hopefully they can let me know why using document.createEvent causes a leak.

petkaantonov:
I don't see how this would work, `createEventObject` is the old proprietary IE6-8 API that works with `attachEvent`, `fireEvent` and `detachEvent`. `createEvent` is DOM2 api that works with `addEventListener`, `dispatchEvent` and `removeEventListener`,.

petkaantonov:
closing as per above

PR #673, opened 2015-06-26
Remove `readmeFilename` from package.json.

ngryman:
Hey,

As per https://github.com/npm/npm/issues/3573#issuecomment-19550501, `readmeFilename` is not official and was droped.
It cannot be ignored by `npm` so it will still be bundled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
     "uglify-js": "~2.4.16",
     "kefir": "^2.4.1"
   },
-  "readmeFilename": "README.md",
   "main": "./js/main/bluebird.js",
   "browser": "./js/browser/bluebird.js",
   "files": [


ngryman:
I don't know why test is failing. I think it's related to `saucelabs` conf.
I see something like this in the logs: `GET https://saucelabs.com/rest/v1/undefined/tunnels?full=1`.

benjamingr:
The CI is shakey, don't worry about it. Are you sure it's not autogenerated? Why is this a big deal?

ngryman:
It's not a big deal at all, it's just cleanup :)

@isaacs explains it in the link I provided. It is not necessary nor standard.

petkaantonov:
merged

PR #674, opened 2015-06-26
LICENSE can't be ignored by npm.

ngryman:
Hey again,

Small cleanup, `LICENSE` does not need to be specified in `files`: https://github.com/npm/npm/blob/master/node_modules/fstream-npm/fstream-npm.js#L91-L124

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
     "js/browser",
     "js/main",
     "js/zalgo",
-    "LICENSE",
     "zalgo.js"
   ]
 }


ngryman:
I don't know why tests are failing. I don't think it's related to my changes, but something with `saucelabs` conf.
I see something like this in the logs: `GET https://saucelabs.com/rest/v1/undefined/tunnels?full=1`.

petkaantonov:
merged

PR #679, opened 2015-06-29
Changed docs to reflect resolve / reject behaviour

chainlink:


API.md
```diff
@@ -267,29 +267,23 @@ class MyCustomError extends Error
...
-predicate function instead of an error constructor, the predicate will receive
-the error as an argument. The return result of the predicate will be used
-determine whether the error handler should be called.
-
-Predicates should allow for very fine grained control over caught errors:
-pattern matching, error-type sets with set operations and many other techniques
-can be implemented on top of them.
-
-Example of using a predicate-based filter:
+If a response is receieved, Request will always resolve successfully. Use statusCode to throw specific errors.
```
ruimarinho: Typo (`received`).

bergus:
Uh, you shouldn't have removed that "*This method also supports predicate-based filters. …"* paragraph - it is still essential to the `catch` documentation! If at all, the `request` example should've been exchanged for some other library, but still demonstrating the same feature of `catch`!

API.md
```diff
@@ -267,29 +267,23 @@ class MyCustomError extends Error
...
 
-function ClientError(e) {
-    return e.code >= 400 && e.code < 500;
+function sensibleRequest(opt) { 
+  return request(opt).spread(function(res, body) {
+    if (response.statusCode >= 300) { 
+      var e = new Error(body); 
+      e.code = response.statusCode; 
+      throw e; 
+    } else {
```
ruimarinho: No need for else since `throw` already breaks the current flow.

petkaantonov:
The PR should only change request to throw on non-successful statuses [similar to what `jQuery.ajax` does](https://github.com/jquery/jquery/blob/master/src/ajax.js#L693). There is no need to change any actual documentation.

benjamingr:
ping @chainlink

chainlink:
>The PR should only change request to throw on non-successful statuses similar to what jQuery.ajax does. There is no need to change any actual documentation.
So, should I modify the library itself to reject the promise on non-successful statuses?

petkaantonov:
No, you changed both the example and the documentation but only the example needs to be changed

petkaantonov:

PR #683, opened 2015-07-01
Design new logo

FranzSkuffka:
I think this library is awesome, much unlike my JS skills. However I still want to contribute and I fancy creating neat logos, so that's what I did here.

## Concept
This design is essentially an evolution of the current logo which preserves some of the present visual identity. However the silhouette of a bluebird is filled with horizontal lines. These parallel lines of different lengths, starting at different points are a metaphor for asynchronously executed tasks.

Take a look at the checkboard for some colors.
I hope you all like it.


bergus:
I have to admit that I like the old logo better, looking like an arrow, curly brace and bird at the same time. And the color scheme is better than boring black :-)
I think the "task schedule" metaphor of the horizontal lines would be better to understand if the long lines had "holes" where they need to wait for others, or the "thread" is sleeping or something.

petkaantonov:
Discussion is in the issue https://github.com/petkaantonov/bluebird/issues/684

PR #687, opened 2015-07-02
Add edge case for iPhone standalone in scheduler which does not fire …

### LINKED ISSUE #666: Promise.resolve not resolving because of MutationObserver (iOS, WebApp)
Debugging from events to Bluebird for a while now on this problem, as it seems to be very environment specific. On iOS 8.3 (iPhone 6 as iPod & iPad last gen.) added a web page to the homescreen, creating a web app produces the following bug.

When you're scrolling in the web app and you let it resolve something based on the `tap` event it will not resolve the promise for 10 seconds (it seems the websocket made it trigger after 10 seconds due to heartbeat). It seems difficult to hit this point at first, but with a web app full of products (40/50) and scrolling trough it, tapping while scrolling is being done very easily. 

The `MutationObserver` will not fire for a while unless you do something else on the webapp. Mostly people will try to tap again, causing the previous promise to resolve while handeling the new one too. The user is suddenly served with its old and new request.

I've build a test app to show what's going on. Hopefully it gives some insides on what's happening. 
http://wehandcraft.com/bluebirdtest/

If you add this page to the homescreen (webapp) in iOS, swipe up (scroll down) and tap a product you'll see `Promise resolve` on the bottom left corner. This means the promise is being resolved, but as you will see, nothing is happening. After a while (if you're lucky) or after a new gesture on the webpage it will say 'resolved'. 

I've left some of the console.logs (haha, yup!) in the `bluebird.js` to give a bit of indication to what I think is the problem. 

Hopefully we'll be able to find a fix for this somehow!

benjamingr:
…MutationObserver changes, closes #666

### DIFF
--- a/src/schedule.js
+++ b/src/schedule.js
@@ -10,7 +10,10 @@ if (util.isNode && typeof MutationObserver === "undefined") {
     schedule = util.isRecentNode
                 ? function(fn) { GlobalSetImmediate.call(global, fn); }
                 : function(fn) { ProcessNextTick.call(process, fn); };
-} else if (typeof MutationObserver !== "undefined") {
+} else if ((typeof MutationObserver !== "undefined") && 
+          !(typeof window !== "undefined" && 
+            window.navigator && 
+            window.navigator.standalone)) {
     schedule = function(fn) {
         var div = document.createElement("div");
         var observer = new MutationObserver(fn);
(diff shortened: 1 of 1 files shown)


bergus:
Ah, docs :-) :+1:

benjamingr:
The CI fails for an authorization error (so not an actual failure) but I'd still feel much easier with a passing CI on different browsers. @petkaantonov anything we can do about it?

PR #692, opened 2015-07-05
Dont recommend use of disposers for transactions (fixes #686)

### LINKED ISSUE #686: Transaction example and using
Transaction commit, sometimes, produces errors (logical errors) because, for example, you are using an optimistic transaction model.

In these cases, the error is not an "internal" error and must be treated as an expected situation (i.e.:  you must return a logical error to the user or retry or ...).

In your examples for Disposing you propose this

```javascript
.disposer(function(client, promise) {
        if (promise.isFulfilled()) {
            return client.queryAsync('COMMIT').then(closeClient);
        } else {
            return client.queryAsync('ROLLBACK').then(closeClient);
        }
        function closeClient() {
            if (close) close(client);
        }
    });
```

Obviously, ROLLBACK must work allways and an error when rollbacking tell us about internall error... 

Problem is COMMIT, because a COMMIT exception must be "logically" treated and the using/disposer pattern causes a crash if disposition fails.... 

A more valid solution (I propose) is taking out disposer the "commit" management and force the caller to perform an explicit commit

```javascript
var pg = require('pg');
Promise.promisifyAll(pg);

function beginTransaction(connectionString) {
    var close;
    var commited=false;
    return pg.connectAsync(connectionString).spread(function(client, done) {
        close = done;
        return client.queryAsync('BEGIN').then(function () {
            return { 
              client: client, 
              commit: function(result){ 
                   return client.queryAsync("COMMIT").then(function(res){ commited=true; return result;});
              }
            };
        });
    }).disposer(function(trx /*, promise*/) {
        if (commited) {
            closeClient();
        } else {
            return trx.client.queryAsync('ROLLBACK').then(closeClient);
        }
        function closeClient() {
            if (close) close(trx.client);
        }
    });
}

exports.beginTransaction = beginTransaction;
```
that can be used in this way:
```javascript
var beginTransaction = require("pghelpers").beginTransaction;
var using = require("bluebird").using;

using( beginTransaction(connectionString), function(trx){
   return trx.client.queryAsync(query).then(trx.commit);
}).then(function(result){
  // valid commited result
}).catch(function(err){
  // if commit fails... we have an error here and we can treat it adequately.
});

Is similar to the old c# "synchronous" pattern for transactions with "using" statements

using(var trans = beginTransaction()){
  ....
  trans.commit();
}

spion:
Here is an updated API.md. I considered adding the real implementation for PG that @ddcovery wrote, but it felt like example overload. Maybe its better to add that to the 3.0 docs only, where every topic has a separate page...

Addresses #686

### DIFF
--- a/API.md
+++ b/API.md
@@ -1335,19 +1335,38 @@ module.exports = getSqlConnection;
 
 The second argument passed to a disposer is the result promise of the using block, which you can inspect synchronously.
 
-Example:
+
+#### Note about disposers in node
+
+If a disposer method throws or returns a rejected promise, its highly likely that it failed to dispose of the resource. In that case, Bluebird has two options - it can either ignore the error and continue with program execution or throw an exception (crashing the process in node.js).
+
+In bluebird we've chosen to do the later because resources are typically scarce. For example, if a database connection cannot be disposed of and Bluebird ignores that, the connection pool will be quickly depleted and the process will become unusable (all requests that query the database will wait forever). Since Bluebird doesn't know how to handle that, the only sensible default is to crash the process. That way, rather than getting a useless process that cannot fulfill more requests, we can swap the faulty worker with a new one letting the OS clean up the resources for us.
+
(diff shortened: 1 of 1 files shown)


petkaantonov:
https://github.com/petkaantonov/bluebird/commit/ddad2317fe93d8164b9e35415094f41a726dd9a3

PR #709, opened 2015-07-22
Fix #708 Promise.map concurrency execuition order issue

### LINKED ISSUE #708: Promise.map concurrency execution order
I am having a problem with the concurrency option for ```Promise.map```. I'm my code I want the order of execution of promises to match the order of the input (in my case optimizing file reads from a CDROM)

Consider this example:
```Javascript
Promise.map([1, 2, 3, 4, 5], function(val) {
    return new Promise.delay(100).then(function() {
        console.log(val);
    });
}, {concurrency: 2});
```
The console output of this code will be ```1 2 5 4 3```

henryqdineen:
See #708. If this isn't acceptable can someone explain why it was implemented the original way?

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -104,7 +104,7 @@ MappingPromiseArray.prototype._drainQueue = function () {
     var values = this._values;
     while (queue.length > 0 && this._inFlight < limit) {
         if (this._isResolved()) return;
-        var index = queue.pop();
+        var index = queue.shift();
         this._promiseFulfilled(values[index], index);
     }
 };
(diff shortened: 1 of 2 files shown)


benjamingr:
Promise.map does not guarantee execution order - use .each or .reduce if you need that.

The order is the way it is as an allowed optimization since the order is not guaranteed

bergus:
@henryqdineen: Surely `pop` is faster than `shift`, because it doesn't have to move every value around.

Issue #720, opened 2015-08-04
Wrapper to Join on all function's arguments


mdarveau:
Hi,

First of all, thank you for this excellent library.

When I have a function that can receive a resolved value or a promise, I often find myself doing the following:
```
exports.doSomething = ( valueOrPromise1, valueOrPromise2 ) ->
  Bluebird.join valueOrPromise1, valueOrPromise2, ( value1, value2 ) ->
    value1.something()
    value2.somethingElse()
```

I finally created a little wrapper that automatically join all my functions params and then call the function that works on resolved values (coffeescript + lodash):
```
joinArgs = (f)->
  _.partialRight(Bluebird.join, f);

exports.doSomething = joinArgs ( value ) ->
  value.something()
```

What do you think of adding the `joinArgs` utility to Bluebird? I can work out the PR but don't want to waste anyone's time if you don't think it belongs to the lib. Maybe joinWrap or wrapJoin would be a better name?

bergus:
`lift` would be an appropriate name.

spion:
FWIW, my experience is that lifted functions can work well locally but the ability to take promises as arguments gets confusing really fast. Built in functions and operators can't do it, and neither can auto-promisified libraries, so you can easily end up with an API that consists of a mix of lifted and unlifted methods. Thats pretty difficult to deal with (unless you use a typed dialect such as typescript/flowtype)

petkaantonov:
Duplicate of https://github.com/petkaantonov/bluebird/issues/78

<hr>

Some things to help figure out how useful addition this would be to core:

1. Do you use this often in your projects? Or just a lot in one project? Or just once in many projects?

2. Are there any other people that would be interested in this?

3. Can it be used as a generic build block that you can build on top of instead of being a specialized tool for some specific scenarios?

bergus:
I'd be interested too. I'd say it's seldom used in many projects - frequency depending on programming style. As @spion says, functions that take promises as arguments (when they don't really require it) are rather confusing in an API. It would be helpful for internal functions though.
I think it's as generic as `Promise.join` as a building block, the only difference is that `Promise.lift` would be curried.

mdarveau:
I only have one node project that uses Promise extensively. However, I use "lift" a lot in services that can be called by different endpoints/resources/context.

[...]

PR #723, opened 2015-08-05
WIP Added 'lift'.

mdarveau:
Following discussion in https://github.com/petkaantonov/bluebird/issues/720.

There are most probably a few loose end. I had a hard time figuring exactly what I had to change in the build file.

Let me know what you think.


src/lift.js
```diff
@@ -0,0 +1,15 @@
...
+module.exports = function(Promise) {
+
+Promise.lift = function ( fn ) {
+    var that = this;
+    return function() {
+        var $_len = arguments.length;
+        var args = new Array($_len+1);
+        for(var $_i = 0; $_i < $_len; ++$_i) {args[$_i] = arguments[$_i];}
+        args[$_len] = fn;
+        return Promise.join.apply( that, args );
```
bergus: Shouldn't this be `that.join.apply(that, args)`?

mdarveau: No, unless we can guarantee that `this`==`Bluebird` when the function is called. I keep `this` in `that` to make sure the original function is called with the same `this` binding than the lifted function.

petkaantonov:

PR #746, opened 2015-08-26
Add short support channel summary in README

spion:
Its short and inviting and right at the beginning to make it harder to miss.


benjamingr:
LGTM

PR #769, opened 2015-09-18
Document unreachable code purpose

madarche:
Document unreachable code purpose since it's triggering "unreachable code after return statement" warning in Firefox when browserify-ied.

cf. https://bugzilla.mozilla.org/show_bug.cgi?id=1151931
cf.  http://stackoverflow.com/questions/24987896/how-does-bluebirds-util-tofastproperties-function-make-an-objects-properties

Note: The text of the comment is a copy of Benjamin Gruenbaum's original comment.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -90,6 +90,7 @@
         "IS_REJECTED": false,
         "IS_CANCELLABLE": false,
         "IS_FINAL": false,
+        "IS_REJECTION_IGNORED": false,
         "IS_BOUND": false,
         "IS_MIGRATED": false,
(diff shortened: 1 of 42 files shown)


benjamingr:
Is there a way to just suppress the warning instead?

madarche:
@benjamingr not to my knowledge after having read all https://bugzilla.mozilla.org/show_bug.cgi?id=1151931 . Nothing like JSHint special comments.

madarche:
Closing in favor of #884 (which is just the same but on current state of the repository).

PR #772, opened 2015-09-20
add explaination for resolve and reject function for beginner

buaya91:
In API.MD, there is not much info about how resolve and reject function are supposed to be used.

### DIFF
--- a/API.md
+++ b/API.md
@@ -93,6 +93,8 @@ Core methods of `Promise` instances and core static methods of the Promise class
 
 Create a new promise. The passed in function will receive functions `resolve` and `reject` as its arguments which can be called to seal the fate of the created promise.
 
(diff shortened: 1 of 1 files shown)


bergus:
Instead of using the terminology "called when failed", I'd rather suggest "called *to fail the promise*". Also, calling `resolve()` doesn't necessarily mean that the promise will complete successfully ("fulfill"), have a look at [What is the correct terminology for javascript promises](http://stackoverflow.com/q/29268569/1048572) on SO.

tflanagan:
This PR doesn't help anyone in understanding promises and the ```resolve, reject``` arguments.

You should use proper promise terminology when explaining the two functions. "Render the outcome" is a confusing and misleading explanation.

buaya91:
Ahh, I see, sorry, I'll go get the terms correct before I submit again

PR #809, opened 2015-10-16
Implement thru method

kjvalencik:
Implements `Promise.prototype.thru`, a convenience method for chaining. This could be implemented by assigning to a variable or wrapping the promise chain in a method, e.g.,

```js
somethingElse(doSomething()
    .then(...)
    .then(...));
```

But, this allows for cleaner chaining, similar to the [`.thru`](https://lodash.com/docs#thru) method in lodash or the [`.through`](http://highlandjs.org/#through) method in highland.js.

```js
doSomething()
    .then()
    .then()
    .thru(somethingElse);
```

### DIFF
--- a/API.md
+++ b/API.md
@@ -62,6 +62,7 @@
     - [`Promise.coroutine.addYieldHandler(function handler)`](#promisecoroutineaddyieldhandlerfunction-handler---void)
 - [Utility](#utility)
     - [`.tap(Function handler)`](#tapfunction-handler---promise)
+    - [`.thru(Function handler)`](#thrufunction-handler---dynamic)
     - [`.call(String propertyName [, dynamic arg...])`](#callstring-propertyname--dynamic-arg---promise)
     - [`.get(String propertyName|int index)`](#getstring-propertynameint-index---promise)
     - [`.return(dynamic value)`](#returndynamic-value---promise)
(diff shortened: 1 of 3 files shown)


shibbybird:
:+1:

benjamingr:
I'm confused - what is this trying to accomplish? Why would you ever want to pass the promise and not its result to a function? If you want to pass the value - that's just `then` anyway.

kjvalencik:
@benjamingr, to allow inlining synchronous operations on a promise instead of waiting for the resolve. Two uses in mind:

* Caching the promise--not the result, so that other calls to the method will wait on the same asynchronous operation.
* Transforming to a stream. Streams are already async, so it doesn't make much sense to resolve a promise to a stream. Example with highland: https://github.com/petkaantonov/bluebird/pull/809/files#diff-88333ba11114e014e87198d04680144dR2242

benjamingr:
Sorry, I just don't see many use cases for this - it's not common enough to be included in core. In general promise returning methods almost never expose the promise and almost always expose the data. You can bind to highland and then `.call` its methods.

Please do feel free to publish it as a separate package though. Those tend to get some downloads and provide features outside of core bluebird. I guarantee that we will not implement a separate `thru` method and that your extension will continue to work forever.

kjvalencik:
Fair enough. I appreciate the feedback. It's a simple two liner and thankfully extending bluebird is simple. 

Cheers!

PR #815, opened 2015-10-26
Nested .get()

### LINKED ISSUE #767: Nested .get()
It would be mighty handy to be able to `.get()` more than one level down. So instead of:

```js
promiseReturningFunc()
  .get('a')
  .get('b')
  .get('c')
  .get('d')
  .then(doSomethingToD)
```

you could use:

```js
promiseReturningFunc()
  .get(['a', 'b', 'c', 'd'])
  .then(doSomethingToD)
```

or even:

```js
promiseReturningFunc()
  .get('a.b.c.d')
  .then(doSomethingToD)
```

like [lodash's _.get()](https://lodash.com/docs#get).

Thanks for making such an awesome library.

dsernst:
Closes #767

This PR adds functionality, documentation, and a unit test for this sort of behavior:

```js
promiseReturningFunc()
  .get(['a', 'b', 'c', 'd'])
  .then(doSomethingToD)
```


dsernst:
This is the only Travis error, appears unrelated:

> 26 Oct 02:48:33 - GET https://saucelabs.com/rest/v1/undefined/tunnels?full=1: HTTP response code indicated failure.26 Oct 02:48:33 - Response: {"error": "Not authorized"}.26 Oct 02:48:33 - Error shutting down overlapping tunnels.26 Oct 02:48:33 - Sauce Connect could not establish a connection.26 Oct 02:48:33 - Please check your firewall and proxy settings.

src/call_get.js
```diff
@@ -106,6 +106,13 @@ function indexedGetter(obj) {
     return obj[index];
 }
 Promise.prototype.get = function (propertyName) {
+    if (util.isArray(propertyName)) {
+        if (propertyName.length > 1) {
+            var firstProperty = propertyName.shift();
+            return this.get(firstProperty).get(propertyName);
```
benjamingr: There should probably be a fast path for where the property is a non-promise.

petkaantonov:
You should probably create a third kind of getter "arrayGetter" and use that as a third case in the if else chain. It will be same amount of code but far more efficient

petkaantonov:
Also, wouldn't multiple args be more comfortable to use than having to create an array:

```js
promiseReturningFunc().get('a', 'b', 'c', 'd')
```

bergus:
> wouldn't multiple args be more comfortable to use than having to create an array

More comfortable for handwritten code, yes, but not when the "path" is dynamically constructed. Imo both should be allowed.
```js
if (arguments.length != 1 || typeof arguments[0] == "object") {
    var path = [].concat.apply([], arguments);
    … // use arrayGetter
} else
    … // fast path for single getter
}
```

petkaantonov:
master was replaced with 3.0, which is divergent from 2.x

petkaantonov:
@bergus I don't think dynamic path is so common, I don't recall ever needing it in application code and in utility code I have only needed dynamic path to implement an utility that parses a string like `a.b.c.d` to a function that reads the property path `a.b.c.d` from an object

petkaantonov:

PR #819, opened 2015-10-27
Fix "home" link in jekyll default layout

hawkrives:
It was linking to `/bluebird/`, but the new site needs it to link to `/`.


petkaantonov:
thanks

PR #822, opened 2015-10-27
Fix 'fromCallback' doc link in 'new in bluebird 3'

fasterthanlime:


fasterthanlime:
( :( saucelabs travis build errored, pretty sure that commit has nothing to do with it )

PR #824, opened 2015-10-27
Fix link on header logo

beeman:
The header logo linked to `/bluebird` which caused a 404. 

This patch links every logo to `/`.


petkaantonov:
Those files are auto-generated from master branch

beeman:
Ah ok. Looks like it's already fixed there :)

PR #830, opened 2015-10-29
docs: Fix "resource management" link

southpolesteve:
Looks like a broken link from how the previous version of the docs were structured


petkaantonov:
[Resource management](.) with dot as the link should automatically expand to an api link

PR #842, opened 2015-11-03
rough intro to bluebird with generators

cultofmetatron:
added a brief overview of generators with information on how to use them with bluebird.

### DIFF
--- a/docs/docs/what-about-generators.md
+++ b/docs/docs/what-about-generators.md
@@ -3,6 +3,15 @@ id: what-about-generators
 title: What About Generators?
 ---
 
+[← Back To API Reference](/docs/api-reference.html)
+<div class="api-code-section"><markdown>
+##Promise.coroutine
+
+```js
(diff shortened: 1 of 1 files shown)


gdi2290:
:+1:

petkaantonov:
The unfinished article tag shouldn't be removed yet. The article is also supposed to explain why generators don't really deprecate promises right now, which is about all the aggregate methods that cannot be replicated, the sucky syntactic catch statement, lack of `using` statement, no long stack traces etc.

docs/docs/what-about-generators.md
```diff
@@ -3,4 +3,63 @@ id: what-about-generators
 title: What About Generators?
 ---
 
-[what-about-generators](unfinished-article)
+[← Back To API Reference](/docs/api-reference.html)
```
petkaantonov: You can link to the actual api documentation by writing

`[Promise.coroutine](.)`

benjamingr:
@cultofmetatron are you still interested in finishing this?

PR #852, opened 2015-11-06
Docs: use Promise.fromCallback instead of Promise.defer().callback

yzarubin:
The docs make use of `defer.callback`, but from running the code and looking at the source code for `Promise.defer` it is apparent that `Promise.defer().callback` is undefined. Instead, use `Promise.fromCallback()`.


benjamingr:
Thanks for the doc fix!

PR #867, opened 2015-11-11
Fix `reduce` docs: omitted initialValue, iterable of length == 1

cspotcode:
The following code demonstrates the behavior I've documented:

```
Promise.resolve(['one']).reduce(function(a, n, i, l) {
    console.log('Arguments are:', a, n, i, l);
    return a + n;
}).then(function(a) {
    console.log('Result is:', a);
});
```
It logs:
```
Result is: one
```


benjamingr:
Thanks, that was embarrassing :D

cspotcode:
No worries, I write Typescript declarations so I'm used to finding tiny
little mistakes in documentation.  :D

PR #868, opened 2015-11-11
Include bluebird in bluebirdjs.com

nickclaw:
Some documentation sites include their own library in the source code so that users can use the debug console as a playground. One example is lodash (https://lodash.com/docs). It makes it a lot easier to quickly test something out. Not a big deal though.


benjamingr:
I'm +1 on this.

petkaantonov:
This is a great idea

PR #869, opened 2015-11-12
Handle promises arrays with properties

lextiz:
In case when variable "values" contains array of promises with properties (http://stackoverflow.com/questions/9952126/add-property-to-javascript-array) creation of new array causes the resolved array not to have the properties of original array (except length).
Proposed solution is copying the custom properties from original array to the resolved array. This solution involves minimal performance overhead (http://jsperf.com/cloning-an-object). I do not use getOwnPropertyNames as it appeared in ECMA script 5.1 only.

### DIFF
--- a/src/promise_array.js
+++ b/src/promise_array.js
@@ -87,7 +87,16 @@ PromiseArray.prototype._init = function init(_, resolveValueIfEmpty) {
 PromiseArray.prototype._iterate = function(values) {
     var len = this.getActualLength(values.length);
     this._length = len;
-    this._values = this.shouldCopyValues() ? new Array(len) : this._values;
+    if (this.shouldCopyValues()) {
+         // Create new array and copy custom properties
+         // getOwnPropertyNames is not used as it appeared in ECMA script 5.1 only
+         this._values = new Array(len);
+         for (var property in values) {
+             if (values.hasOwnProperty(property)) {
+                 this._values[property] = values[property];
+             }
+         }
+    }
     var result = this._promise;
     var isResolved = false;
     var bitField = null;
--- a/src/promise_array.js
+++ b/src/promise_array.js
@@ -88,8 +88,6 @@ PromiseArray.prototype._iterate = function(values) {
     var len = this.getActualLength(values.length);
     this._length = len;
     if (this.shouldCopyValues()) {
-         // Create new array and copy custom properties
-         // getOwnPropertyNames is not used as it appeared in ECMA script 5.1 only
          this._values = new Array(len);
          for (var property in values) {
              if (values.hasOwnProperty(property)) {


phpnode:
Strongly disagree with this change, this method produces a new array and there should be no expectation that ad hoc properties should be carried over too. And what should the behaviour be if the properties are promises? should those be resolved too? I don't think this is a good idea.

lextiz:
Why there should be no expectation that custom properties are persisted? As, no specification talks about this topic the decision is on implementors.
I think such behaviour should be supported because:
1) Adding properties to arrays is legitimate behaviour in JS. Person who adds properties to promises array naturally expects to find these properties on values array. This behaviour is more intuitive for users of library.
2) Such behaviour is accepted in another promises libraries (e.g. Q)

Properties that are promises should not be handled, as far as I understand resolving array of promises means resolving each element in the array, not the properties of it.

In general for bluebird what is the priority in tradeoffs between usability of framework to users and simplicity of the library code?

Anyway in case the decision is to leave the existing behaviour I think it should be explicitly stated in APIs description that there is no connection between promises array and the values array, when resolved.

lextiz:
One more thing to clarify what I mean as "not intuitive behaviour": if the library offers "seamless conversion of promises array to values array" functionality and accepts arrays created by user than the user can expect that the array properties to stay. To understand that the arrays are actually different user has to be familiar with library implementation.

phpnode:
If I extend array, like this:
```js
class UserCollection extends Array {}
```
and pass it to Bluebird, should Bluebird give me back an array or an instance of `UserCollection`?

lextiz:
If we talk about ECMAScript 6 enabled environments, then clone function should be used to clone the whole object (original UserCollection array) first and instance of UserCollection should be returned. Do you have ES6 shim included into the library?

phpnode:
> then clone function should be used to clone the whole object (original UserCollection array) first and instance of UserCollection should be returned

That's why this is a mistake. Just because we were passed an array of one kind, does not mean that it makes sense for Bluebird to return an array of that kind. For one, it has no idea how to instantiate a `UserCollection`, just like it has no idea whether the properties which happened to be on the input array of promises make sense for an array of values. It's magic behaviour and confusing for users, because the output array is distinct and separate from the input, and that fact needs to be clear.

```js
const input = [1,2,3];
input.foo = 'bar';

output = input.map(item => item);

console.log(output.foo); // what does the user expect here?
```

benjamingr:
Bluebird should behave like native promises and call the constructor of whatever is passed in to the array. I remember talks of @@species but honestly I'm not that well versed. Perhaps @domenic can elaborate on the correct expected promise behavior?

petkaantonov:
Bluebird doesn't generally follow ES6 on the really subtle subclassing stuff anyway.

It doesn't even work in Chrome 46:

```
Promise.all(new SubArray(1,2,3)).then(function(v) {
     console.log(v instanceof SubArray)
});
VM1444:3 false
Promise {[[PromiseStatus]]: "pending", [[PromiseValue]]: undefined}
```

petkaantonov:
Also Q doesn't do this, Q returns literally the original input array that's why it still has the properties. But that approach has other obvious problems...

bergus:
@lextiz 
> Adding properties to arrays is legitimate behaviour in JS.

Well, it's possible of course because arrays are objects, but it's a [despised practise](http://andrewdupont.net/2006/05/18/javascript-associative-arrays-considered-harmful/). What are you using these properties for?

> Person who adds properties to promises array naturally expects to find these properties on values array.

No, it's clear to everyone that `Promise.all` resolves with a *new* array. If you must use some custom properties, you can still refer to the property on the old array without any problems.
If you are doing some kind of subclassing of `Array`, creating a new subclass instance (via ES6 `Symbol.species` etc) might be feasible, but just blindly copying over properties is no good.

benjamingr:
Or you can just use `Promise.props`...

Issue #566, opened 2015-04-07
Build fails on windows with CRLF line endings 


HenriKarapuu:
tools/build.js throws if git or any editor changes the line endings to CRLF (which they often do by default on Windows). This makes working on bluebird codebase a hassle, as every tool and editor needs to be manually configured to use LF line endings just for bluebird. Should be simple to fix?

Unhandled rejection Error: The file F:\Projects\_Modified\JavaScript\bluebird\src\bind.js
contains an illegal character:
 (U+000D) on line 1 at column 14
    at noStackError (F:\Projects\_Modified\JavaScript\bluebird\tools\utils.js:11:13)
    at F:\Projects\_Modified\JavaScript\bluebird\tools\utils.js:26:23
    at Array.forEach (native)
    at Object.checkAscii (F:\Projects\_Modified\JavaScript\bluebird\tools\utils.js:18:30)
    at eval (eval at getTaskFunction (F:\Projects\_Modified\JavaScript\bluebird\tools\job-runner\job-runner.js:413:33), <anonymous>:5:27)
    at eval (eval at getTaskFunction (F:\Projects\_Modified\JavaScript\bluebird\tools\job-runner\job-runner.js:413:33), <anonymous>:4:31)
    at promise.then.finally.then.process.send.type (F:\Projects\_Modified\JavaScript\bluebird\tools\job-runner\job-runner.js:507:32)
    at process._tickCallback (node.js:419:13)
    at Object.actions.newTask (F:\Projects\_Modified\JavaScript\bluebird\tools\job-runner\job-runner.js:500:22)
    at Object._onImmediate (F:\Projects\_Modified\JavaScript\bluebird\tools\job-runner\job-runner.js:546:39)
    at processImmediate [as _immediateCallback] (timers.js:345:15)
From previous event:
    at Object.ret.run (F:\Projects\_Modified\JavaScript\bluebird\tools\job-runner\job-runner.js:141:27)
    at F:\Projects\_Modified\JavaScript\bluebird\tools\build.js:317:30
From previous event:
    at F:\Projects\_Modified\JavaScript\bluebird\tools\build.js:316:32
    at process._tickCallback (node.js:419:13)
From previous event:
    at Module.build [as exports] (F:\Projects\_Modified\JavaScript\bluebird\tools\build.js:315:20)
    at Object.<anonymous> (F:\Projects\_Modified\JavaScript\bluebird\tools\build.js:365:12)
    at Module._compile (module.js:456:26)
[...]

PR #872, opened 2015-11-17
Fixed build on windows

lextiz:
"npm install" on windows failed with unhandled rejection error, which was caused by os specific line separator used for parsing source files.

### DIFF
--- a/tools/utils.js
+++ b/tools/utils.js
@@ -4,9 +4,8 @@ var path = require("path");
 var spawn = require("cross-spawn");
 Promise.longStackTraces();
 var fs = Promise.promisifyAll(require("fs"));
-var notAscii = /[^\u0019-\u007E]/;
+var notAscii = /[^\u000D\u0019-\u007E]/;
 var Table = require('cli-table');
-var os = require('os');
 
 function noStackError(message) {
     var e = new Error(message);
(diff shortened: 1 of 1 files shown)


petkaantonov:
What if "\n" is used in Windows as well?

lextiz:
This case is not handled in proposed solution. Actually the line endings in Windows depend on Git configuration, and as far as I understand the default configuration for Windows is CRLF in most cases (at least for Git-Bash, Cygwin and Egit). Is there some real use case when line endings in Windows would be \n? We could also remove \r from notAscii characters array, as it is actually ASCII character, does it sound more convenient to you?

petkaantonov:
Actually I would ask if there is a real use case for using \r\n?

lextiz:
Yes.
1) Download git for Windows from official site and install it
2) Clone bluebird repo
3) run "npm install" (assuming latest node.js is installed and in path)
4) build.js will fail with exception

petkaantonov:
I don't have a fresh git for windows install but it works just fine

lextiz:
Git for Windows has 3 options regarding line endings handling in its installer and "Checkout Windows-style, commit Unix-style line endings" is the default one. You have probably changed this setting. Is the option moving \r from notAscii characters array acceptable? This option would work in all cases.

petkaantonov:
But currently the code splits by \r\n on Windows which doesn't work when \n is used in Windows (for example I use \n in Windows)

lextiz:
I have submitted the alternative solution: added \r to notAscii regexp.

petkaantonov:
Thanks

bergus:
Hm, I thought this should've been fixed via #572 (after #566). So even on Windows you should have Unix line endings after checking out.

PR #883, opened 2015-11-22
Travis script updates

vitaly-t:
* Set testing against environments that matter today - NodeJS 0.10, 0.12, 4.x and 5.x
* Moved the same environment settings into their own section
* Added support for the new Travis CI containment mode: `sudo: false`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,12 @@ sudo: false
 matrix:
   include:
     - node_js: "0.10"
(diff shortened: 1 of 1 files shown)


petkaantonov:
thanks

vitaly-t:
You are welcome! And since you are the account owner, I think there must be a way for you to re-run checks on all the PR-s that failed before ;)

PR #884, opened 2015-11-23
Document unreachable code purpose

madarche:
This is the same PR as #769 but on the current state of the repository because I didn't manage to rebase my previous commit.

Document unreachable code purpose since it's triggering "unreachable code after return statement" warning in Firefox when browserify-ied.

cf. https://bugzilla.mozilla.org/show_bug.cgi?id=1151931
cf. http://stackoverflow.com/questions/24987896/how-does-bluebirds-util-tofastproperties-function-make-an-objects-properties

Note: The text of the comment is a copy of Benjamin Gruenbaum's original comment.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -201,6 +201,9 @@ function toFastProperties(obj) {
     while (l--) new FakeConstructor();
     ASSERT("%HasFastProperties", true, obj);
     return obj;
+    // Prevent the function from being optimized through dead code elimination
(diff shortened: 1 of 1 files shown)


nazrhyn:
:+1:

bdkent:
This information should be on the main website too (like http://bluebirdjs.com/docs/install.html). Having the information just in the source code doesn't help the new bluebird user who installed and started getting warnings in Firefox.

PR #885, opened 2015-11-23
Add padding to the bottom of pages

kimmobrunfeldt:
Makes documentation pages visually a bit more pleasing. I didn't succeed to get jekyll running so didn't actually test this locally.


spion:
Yep, previously you'd get that "why did scrolling break? oh its the end of the page" feel - the 50px padding seems to take care of that.

PR #887, opened 2015-11-23
Added steed to benchmarks

mcollina:
I've added [steed](http://npm.im/steed) to the benchmarks.

It takes a different approach compared to neo-async, and possibly the performance might be different. It recycles functions, so they can be optimized (vs allocating new functions). In my tests, performance is similar or better (and it does not pollute `--trace_opt`): here it comes slightly lower, and I still have to investigate what is happening.

Anyway, it is worth benchmarking.

BTW, I have found that the benchmarks are more favorable to promise-based apps, because they allocates a lot of functions, which in most cases are not needed in a callback-based implementation.


vitaly-t:
#### Noble Steed? LOL :) :) :)


Noble Steed

mcollina:
I've added a couple more benchmarks that are more optimized for callbacks. On my box:

```
file                                      time(ms)  memory(MB)
callbacks-baseline.js                          337       49.33
callbacks-mcollina-fastparallel.js             391       71.87
promises-bluebird-generator.js                 407       78.14
promises-bluebird.js                           461       74.73
callbacks-mcollina-steed-map.js                488       72.15
callbacks-mcollina-steed-parallel.js           545      116.71
callbacks-suguru03-neo-async-parallel.js       553      100.25
```

I originally thought to name my lib [ludicrous](https://www.youtube.com/watch?v=ygE01sOhzz0) :D.

(The picture is copyright by Disney I presume).

benjamingr:
@spion can you please review as this is benchmark code?

benchmark/doxbee-sequential-errors/callbacks-mcollina-steed-waterfall.js
```diff
@@ -0,0 +1,71 @@
+require('../lib/fakes');
+var steed = require('steed');
+
+module.exports = function upload(stream, idOrPath, tag, done) {
+    var blob = blobManager.create(account);
+    var tx = db.begin();
+    var blobId, file, version, fileId;
+    async.waterfall([
```
spion: Seems like this is still trying to use async instead of steed?

benjamingr: lol ^_^

mcollina: Fixed :D.

spion:
Looks okay to me, but I haven't had a chance to test whether everything executes correctly. The trouble with the benchmark code is that there is no verification that all implementations ultimately do the same thing...

benjamingr:
Thanks, I think we can merge - @petkaantonov ?

benjamingr:
@petkaantonov

PR #889, opened 2015-11-25
Update catchthrow.md

SDSWanderer:
Fix catchThrow sample


petkaantonov:
That's not what `catchThrow` does, it throws the `reason` you pass as argument to it, not the `reason` the promise was rejected with. Actually it would be a completely useless method if it did `catch(reason => throw reason)` as that is already the default.

Issue #897, opened 2015-12-01
Bluebird 3.x performance in webpack


suprememoocow:
This issue seems totally bizarre to me. I'm not sure whether to file it under webpack or Bluebird, but since it may be affecting other bluebird users I'll file it here.

I've been finding that Bluebird's performance in the browser has been pretty shoddy since I upgraded to bluebird 3. 

In an effort to diagnose the problem, I started conducting some tests here: https://github.com/suprememoocow/bluebird-webpack-perf

The results are strange to say the least, but here's some of what I'm seeing:

* A simple `Promise.reduce`  with an array of 100 values takes 2ms when run in Chrome 64bit on OSX. The exact same code, when processed through webpack takes 500ms. 

* You can compare the implementations here:
https://github.com/suprememoocow/bluebird-webpack-perf/blob/master/bluebird-no-webpack.html
and
https://github.com/suprememoocow/bluebird-webpack-perf/blob/master/bluebird-with-webpack.js

* In an effort to isolate the problem, I've minimized the processing that webpack does with the following settings:

```js
      module: {
        noParse: [
          /bluebird\/js\/browser\/bluebird\.js/
        ]
      },
      node: {
        console: false,
        global: false,
        process: false,
        Buffer: false,
        __filename: false,
        __dirname: false,
        setImmediate: false
      },
```

* Changing these settings hasn't helped.

* Downgrading to Bluebird 2.x.x fixes the problem.

I'll conduct some further tests and see if I can get to the bottom of the problem

suprememoocow:
More strange results:

The problem **does not** occur in Firefox 42.0 or Safari 9.0.1 on OSX. It is occurring in Chrome 47.0.2526.73 beta (64-bit) and Chrome Canary Version 49.0.2578.0 canary (64-bit)

suprememoocow:
Doing some profiling I noticed that bluebird is capturing long stack traces even though I have 
```js
Promise.config({
    warnings: false,
    longStackTraces: false, // <-- Turn off long stack traces
    cancellation: true
  });
```
This would explain why Chrome is getting the performance hit, but not Firefox or Safari.

Upon further investigation, it looks like the `__DEBUG__` constant is being set to `true` for the build of `js/browser/bluebird.js` if you compare:

```
js/browser/bluebird.js:658:var debugging =!!(true || util.env("BLUEBIRD_DEBUG") ||
```

[...]

PR #898, opened 2015-12-01
It's not currently possible to disable long stack traces when using Bluebird via webpack or browserify

suprememoocow:
This PR fixes the issue described in https://github.com/petkaantonov/bluebird/issues/897

When using Bluebird with webpack or browserify, the `./js/browser/bluebird.js` file is used, which is currently build with `__DEBUG__` set to `true`.

This has a major performance impact on Bluebird in a production environment.

This PR sets `__DEBUG__` to false by default, except if `build.js` is invoked with `--debug`.

Please let me know if there are any changes you would like me to make.

### DIFF
--- a/tools/build.js
+++ b/tools/build.js
@@ -186,7 +186,7 @@ function buildDebug(sources, depsRequireCode, dir) {
     });
 }
 
-function buildBrowser(sources, dir, tmpDir, depsRequireCode, minify, npmPackage, license) {
+function buildBrowser(sources, dir, tmpDir, depsRequireCode, minify, enableDebug, npmPackage, license) {
     return Promise.join(dir, tmpDir, npmPackage, license, function(dir, tmpDir, npmPackage, license) {
         return Promise.map(sources, function(source) {
             return jobRunner.run(function() {
@@ -229,8 +229,9 @@ function buildBrowser(sources, dir, tmpDir, depsRequireCode, minify, npmPackage,
                     src = src + alias;
                     src = src.replace(/\brequire\b/g, "_dereq_");
                     var minWrite, write;
+                    var debugString = enableDebug ? "true" : "false";
                     if (minify) {
-                        var minSrc = src.replace( /__DEBUG__/g, "false" );
+                        var minSrc = src.replace( /__DEBUG__/g, debugString );
                         minSrc = UglifyJS.minify(minSrc, {
                             comments: false,
                             compress: true,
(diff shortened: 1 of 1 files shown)


malditogeek:
:+1:

spion:
Is there any way to enable `__DEBUG__` when e.g. `browserify` is invoked with `--debug` ?

suprememoocow:
@spion, you can run `node tools/build.js --debug --release --browser --minify` to have `__DEBUG__` set to true to get a __DEBUG__ build, but you can't set it from `browserify` since the built artifacts are bundled with the npm package and accessed via the `  "browser": "./js/browser/bluebird.js"` setting in bluebird's `package.json`, not generated on the fly

spion:
@suprememoocow sure; but one of the things that bluebird makes easy is getting started with it: the defaults are developer-friendly (long stack traces when NODE_ENV=development and in the browser, logging possibly unhandled rejections...)

So I was wondering if it may be possible to both keep that property and not require custom builds. (e.g. if browserify injected a `process.browserify.debug = true` global, we could use that...

suprememoocow:
Thats a great point, but I'm not sure whether it's in the scope of this pull-request as there are additional things to take into account if you wanted to allow that:

For example, relying on `process.browserify.debug` would break in webpack. You could possibly use `process.env` which webpack and browserify will shim for you but it would need further discussion.

This PR is just making the `browser` build consistent with other builds:

#### Current situation

| Build | Long Stack Traces | Possible to disable long stack traces? | 
|--------|---------------------------|--------------------------------------------------|
| node | Off by default, can be enabled via env vars | Yes |
| browser minified | Off by default | Yes |
| browserify/webpack | **On by default** | **No** |

#### This PR

| Build | Long Stack Traces | Possible to disable long stack traces? | 
|--------|---------------------------|--------------------------------------------------|
| node | Off by default, can be enabled via env vars | Yes |
| browser minified | Off by default | Yes |
| browserify/webpack | **Off by default** | **Yes** |

spion:
Technically, we could fix things so you *can* turn off LST by using e.g. envify with browserify.

    browserify index.js -t [ envify --BLUEBIRD_LONG_STACK_TRACES 0 ]

Right now that doesn't work due to the fact that a helper is used to read environment vars.

spion:
Ah, now I see: "Possible to disable long stack traces?" - "No". Yeah, thats definitely a real problem.

suprememoocow:
Yeah, perhaps I haven't been explaining the problem in the right way :)

petkaantonov:
So the only problem is `longStackTraces: false` not disabling long stack traces? That should be fixed instead I think

suprememoocow:
@petkaantonov in the webpack build, yes, that's correct. Can you suggest another approach?

petkaantonov:
It's not straight forward so I just did it myself

suprememoocow:
Thanks for the quick response @petkaantonov

PR #900, opened 2015-12-03
Add a first draft of an Async Dialog Tutorial

sukima:
This is an attempt to show Bluebird and Promises when used as values from a dialog box. I use only Bluebird as a dependency and the DOM code is vanilla JS not jQuery.

I first compare the `window.confirm` synchronous code with a asynchronous dialog written in the DOM. Then I refactor the example to a few OOP abstractions which allow be to demonstrate different types of dialogs like a prompt.

I follow this with an explanation on how to best handle the cancellation of the dialog box using custom Errors.

And finally I offer and example on how to provide progress updates by using a progress bar dialog.

### DIFF
--- a/docs/docs/async-dialogs.md
+++ b/docs/docs/async-dialogs.md
@@ -4,3 +4,628 @@ title: Async Dialogs
 ---
 
 [async-dialogs](unfinished-article)
+
+Typically *promises* are used in conjunction with asynchronous tasks such as a
+network request or a `setTimeout`; a lesser explored use is in dealing with user
+input. Since a program has to wait for a user to continue some actions it makes
+sense to consider it an asynchronous event.
+
+For comparison I'll start with an example of a *synchronous* user interaction
+using the `window.prompt` and then move to an *asynchronous* interaction my
+making our own DOM based prompt. To begin here is a template for a simple HTML
+page:
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Async Dislogs Example</title>
+  <script src="//cdn.jsdelivr.net/bluebird/3.0.5/bluebird.js"></script>
+  <script type="text/javascript">
+    document.addEventListener('DOMContentLoaded', function() {
+      var time = document.getElementById('time-stamp');
+      clockTick();
+      setInterval(clockTick, 1000);
+      function clockTick() {
+        time.innerHTML = '' + new Date();
+      }
+    });
+  </script>
+</head>
+<body>
+  <p>The current time is <span id="time-stamp"></span>.</p>
+  <p>Your name is <span id="prompt"></span>.</p>
+  <button id="action">Set Name</button>
+</body>
+</html>
+```
+
+`window.prompt` blocks the web page from processing while it waits for the user
+to enter in data. I has to block because the input is returned and the next line
+of code needs that result. But for sake of this tutorial we are going to convert
+the typical conditional code into a promise API.
+
+```javascript
+function promptPromise(message) {
+  return new Promise(function(resolve, reject) {
+    var result = window.prompt(message);
+    if (result != null) {
+      resolve(result);
+    } else {
+      reject();
+    }
+  });
+}
+
+document.addEventListener('DOMContentLoaded', function() {
+  var button = document.getElementById('action');
+  var output = document.getElementById('prompt');
+
+  button.addEventListener('click', function() {
+    promptPromise('What is your name?')
+      .then(function(name) {
+        output.innerHTML = '' + name;
+      })
+      .catch(function() {
+        output.innerHTML = '¯\\_(ツ)_/¯';
+      });
+  });
+});
+```
+
+[Run example on JSBin](http://jsbin.com/kowama/edit?js,output)
+
+This doesn't add much much using `window.prompt`; however, one advantage is the
+API that promises provide. In the case where we call `promisePrompt(…)` we can
+easily react to the result of the dialog without having to worry about how it is
+implemented. In our example we've implemented the `window.prompt` but our call
+to `promisePrompt()` doesn't care. This makes a change to an *asynchronous*
+dialog a little more future proof.
+
+To drive home the synchronous nature of the `window.prompt` notice that the time
+stops ticking when the prompt dialog is displayed. Lets fix that by making our
+own prompt. Since our dialog is just DOM manipulation the page won't be blocked
+while waiting for user input.
+
+First add the prompt dialog to the HTML:
+
+```html
+<style type="text/css">
+  #dialog {
+    width:      200px;
+    margin:     auto;
+    padding:    10px;
+    border:     thin solid black;
+    background: lightgreen;
+  }
+  .hidden {
+    display: none;
+  }
+</style>
+<div id="dialog" class="hidden">
+  <div class="message">foobar</div>
+  <input type="text">
+  <div>
+    <button class="ok">Ok</button>
+    <button class="cancel">Cancel</button>
+  </div>
+</div>
+```
+
+We will want to keep the same API so our change will be only to the
+`promisePrompt`. It will find the dialog DOM elements, attach events to the
+elements, show the dialog box, return a promise that is resolved based on the
+attached events, and finally detaches the events and cleans up after itself
+(hiding the dialog box for another use later).
+
+```javascript
+function promptPromise(message) {
+  var dialog       = document.getElementById('dialog');
+  var input        = dialog.querySelector('input');
+  var okButton     = dialog.querySelector('button.ok');
+  var cancelButton = dialog.querySelector('button.cancel');
+  var resolver, rejector;
+
+  dialog.querySelector('.message').innerHTML = '' + message;
+  dialog.className = '';
+
+  function handleOk() {
+    resolver(input.value);
+  }
+
+  function handleCancel() {
+    rejector();
+  }
+
+  okButton.addEventListener('click', handleOk);
+  cancelButton.addEventListener('click', handleCancel);
+
+  return new Promise(function(resolve, reject) {
+    resolver = resolve;
+    rejector = reject;
+  })
+  .finally(function() {
+    okButton.removeEventListener('click', handleOk);
+    cancelButton.removeEventListener('click', handleCancel);
+    dialog.className = 'hidden';
+  });
+}
+```
+
+[Run example on JSBin](http://jsbin.com/fucofu/edit?js,output)
+
+Now when the user presses the **Set Name** button the clock continues to update
+while the dialog is visible.
+
+Because the `removeEventListener` requires a reference to the original function
+that was used with the `addEventListener` it makes it difficult to clean up
+after itself without storing the references in a scope higher then the promise
+resolution function. This pattern of storing the `resolve` and `reject`
+callbacks outside the scope of the promise function is called [the **deferred**
+anti-pattern](api/anti-patterns.html#the-deferred-anti-pattern). It is an
+*anti-pattern* because it is often used when it is not needed. However, in this
+case it is the only way to accomplish the clean-up requirements imposed by the
+DOM API (namely the need for the original function reference to
+`removeEventListener`).
+
+If this was using jQuery there would be no need for the deferred pattern because
+jQuery can remove events handlers without the need for a function reference
+using [event namespacing](https://api.jquery.com/on/#event-names).
+
+```javascript
+return new Promise(function(resolve, reject) {
+  $('#okButton').on('click.promptDialog', function() {
+    resolve(input.value);
+  });
+  $('#cancelButton').on('click.promptDialog', reject);
+})
+.finally(function() {
+  $('#okButton').off('click.promptDialog');
+  $('#cancelButton').off('click.promptDialog');
+});
+```
+
+Now looking at the example above there are a few issues with it. Mainly the need
+for the deferred anti-pattern and it feels like it is doing too much. A *squint*
+test reveals four inner functions, six variable references, and its behavior is
+to show the dialog, attach two DOM events, construct a promise, hide the dialog,
+and finally detach DOM events. That is a lot for one little function. A
+refactoring can help.
+
+Abstraction is the key here. We will make an *object* (or class) that is
+responsible for managing the dialog box. Its interface will manage only two
+function references (callbacks): when the user clicks ok and when user clicks
+cancel. And it will offer the value when asked.
+
+Using an abstraction like this the `promisePrompt` no longer needs to know
+anything about the DOM and concentrates on just providing a promise. This will
+also make things easier to create a promised version of a progress bar or
+confirmation dialog or any other type of UI that we want to have a value for.
+All we will need to do is write a class for that dialog type with the same
+interface and just pass that class into our promise making method.
+
+The dialog interface might look like this:
+
+```javascript
+var noop = function() {
+  return this;
+};
+
+function Dialog() {
+  this.setCallbacks(noop, noop);
+}
+Dialog.prototype.setCallbacks = function(okCallback, cancelCallback) {
+  this._okCallback     = okCallback;
+  this._cancelCallback = cancelCallback;
+  return this;
+};
+Dialog.prototype.waitForUser = function() {
+  var _this = this;
+  return new Promise(function(resolve, reject) {
+    _this.setCallbacks(resolve, reject);
+  });
+};
+Dialog.prototype.show = noop;
+Dialog.prototype.hide = noop;
+```
+
+Initially the Dialog class sets the two callbacks to *noop* functions. It is up
+to the child class to call them when necessary. We break down the promise
+creation to one function `waitForUser()` that sets the callbacks and returns a
+promise. At this level the `show()` and `hide()` are just *noop* functions as
+well and will be implemented by the child classes.
+
+Our `PromptDialog` class is responsible for inheriting from `Dialog` and setting
+up the required DOM scaffolding and eventually call `this._okCallback` or
+`this._cancelCallback` as appropriate.
+
+It might look like this:
+
+```javascript
+function PromptDialog() {
+  Dialog.call(this);
+  this.el           = document.getElementById('dialog');
+  this.inputEl      = this.el.querySelector('input');
+  this.messageEl    = this.el.querySelector('.message');
+  this.okButton     = this.el.querySelector('button.ok');
+  this.cancelButton = this.el.querySelector('button.cancel');
+  this.attachDomEvents();
+}
+PromptDialog.prototype = Object.create(Dialog.prototype);
+PromptDialog.prototype.attachDomEvents = function() {
+  var _this = this;
+  this.okButton.addEventListener('click', function() {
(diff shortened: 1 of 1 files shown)


docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
 ---
 
 [async-dialogs](unfinished-article)
+
+Typically *promises* are used in conjunction with asynchronous tasks such as a
+network request or a `setTimeout`; a lesser explored use is in dealing with user
```
benjamingr: Is "in" correct here?

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+input. Since a program has to wait for a user to continue some actions it makes
+sense to consider it an asynchronous event.
+
+For comparison I'll start with an example of a *synchronous* user interaction
+using the `window.prompt` and then move to an *asynchronous* interaction my
```
benjamingr: I don't think `the` is correct here - just "using `window.prompt`" sounds more correct.

benjamingr: Also, should by `by` and not `my`

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+making our own DOM based prompt. To begin here is a template for a simple HTML
```
benjamingr: Need comma after `begin`?

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+page:
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Async Dislogs Example</title>
+  <script src="//cdn.jsdelivr.net/bluebird/3.0.5/bluebird.js"></script>
```
benjamingr: This should probably be the latest version of bluebird. a JSFiddle link might also be nice (an embedded snippet maybe?)

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+  <p>The current time is <span id="time-stamp"></span>.</p>
+  <p>Your name is <span id="prompt"></span>.</p>
+  <button id="action">Set Name</button>
+</body>
+</html>
+```
+
+`window.prompt` blocks the web page from processing while it waits for the user
+to enter in data. I has to block because the input is returned and the next line
+of code needs that result. But for sake of this tutorial we are going to convert
```
benjamingr: Needs a link to the promise constructor or promisification here somewhere for the below example.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+the typical conditional code into a promise API.
+
+```javascript
+function promptPromise(message) {
+  return new Promise(function(resolve, reject) {
+    var result = window.prompt(message);
+    if (result != null) {
+      resolve(result);
+    } else {
+      reject();
```
benjamingr: Always reject with an `Error` instance, never with `undefined`.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+    }
+  });
+}
+
+document.addEventListener('DOMContentLoaded', function() {
```
benjamingr: Why do we need a `DOMContentLoaded` event here?

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+  }
+
+  function handleCancel() {
+    rejector();
+  }
+
+  okButton.addEventListener('click', handleOk);
+  cancelButton.addEventListener('click', handleCancel);
+
+  return new Promise(function(resolve, reject) {
```
benjamingr: Please put the handlers in the promise constructor rather than extracting them outside.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+
+```javascript
+function promptPromise(message) {
+  var dialog       = document.getElementById('dialog');
+  var input        = dialog.querySelector('input');
+  var okButton     = dialog.querySelector('button.ok');
+  var cancelButton = dialog.querySelector('button.cancel');
+  var resolver, rejector;
+
+  dialog.querySelector('.message').innerHTML = '' + message;
```
benjamingr: Why the `'' + message` vs `message`?

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+Because the `removeEventListener` requires a reference to the original function
+that was used with the `addEventListener` it makes it difficult to clean up
+after itself without storing the references in a scope higher then the promise
+resolution function. This pattern of storing the `resolve` and `reject`
+callbacks outside the scope of the promise function is called [the **deferred**
+anti-pattern](api/anti-patterns.html#the-deferred-anti-pattern). It is an
+*anti-pattern* because it is often used when it is not needed. However, in this
+case it is the only way to accomplish the clean-up requirements imposed by the
+DOM API (namely the need for the original function reference to
+`removeEventListener`).
```
benjamingr: That's not really true. The deferred anti pattern is about explicit construction (not extracting them outside), you can just put them in named functions inside the promise construction. See - stackoverflow.com/questions/23803743/what-is-the-explicit-promise-construction-antipattern-and-how-do-i-avoid-it 

This is just promisifying an API (you needed the promise constructor) which you need to do because it's not in a standard format.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+cancel. And it will offer the value when asked.
+
+Using an abstraction like this the `promisePrompt` no longer needs to know
+anything about the DOM and concentrates on just providing a promise. This will
+also make things easier to create a promised version of a progress bar or
+confirmation dialog or any other type of UI that we want to have a value for.
+All we will need to do is write a class for that dialog type with the same
+interface and just pass that class into our promise making method.
+
+The dialog interface might look like this:
```
benjamingr: I think it's starting to lose a little focus here. You're introducing a class without an obvious immediate benefit.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+
+Our `PromptDialog` class is responsible for inheriting from `Dialog` and setting
+up the required DOM scaffolding and eventually call `this._okCallback` or
+`this._cancelCallback` as appropriate.
+
+It might look like this:
+
+```javascript
+function PromptDialog() {
+  Dialog.call(this);
```
benjamingr: Use `class`es for showing classes, not functions :)

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+  this.el           = document.getElementById('dialog');
+  this.inputEl      = this.el.querySelector('input');
+  this.messageEl    = this.el.querySelector('.message');
+  this.okButton     = this.el.querySelector('button.ok');
+  this.cancelButton = this.el.querySelector('button.cancel');
```
benjamingr: For what it's worth - this is a terrible pattern in real code, classes that manage dom nodes should get them as a parameter to the constructor (dependency injection) rather than query the global view state for them.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+})
+.catch(ArgumentError, function(e) {
+  console.log('You buggered up something with the arguments.', e);
+})
+.catch(SyntaxError, function(e) {
+  console.log('Check your syntax!', e);
+})
+.catch(function(e) {
+  // e is an Error object.
+  console.log('Well something genaric happened.', e);
```
benjamingr: It might be a good idea to `throw e;` here as well to let unhandled rejection tracking show you the big stack trace.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+
+To do this we will have two `catch()` functions one for `UserCanceledError` and
+one for any other `Error`. We can make a custom error like so:
+
+```javascript
+function UserCanceledError() {
+  this.name = 'UserCanceledError';
+  this.message = 'Dialog cancelled';
+}
+UserCanceledError.prototype = Object.create(Error.prototype);
```
benjamingr: There is a `makeError` helper the library comes with.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+    _this.cancel();
+  });
+};
+```
+
+And in our usage case we can test for it:
+
+```javascript
+// Timeout the dialog in five seconds.
+setTimeout(function() { prompt.cancel(); }, 5000);
```
benjamingr: Prefer `Promise.delay`, or `Promise.timeout`

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+```
+
+See [this StackOverflow answer](http://stackoverflow.com/a/17891099/227176) for
+a more detailed and feature complete way to make custom errors.
+
+Now we can add a `cancel()` reject with this in our event listener:
+
+```javascript
+Dialog.prototype.cancel = function() {
+  this._cancelCallback(new UserCanceledError());
```
benjamingr: I still don't understand why all these private variables and state are required here. Also, cancellation is done with cancellation - and not rejections.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+    prompt.hide();
+  });
+```
+
+[Run example on JSBin](http://jsbin.com/yaropo/edit?js,output)
+
+**NOTE:** Bluebird supports [cancellation](api/cancellation.html) as an optional
+feature that is turned off by default. However, its implementation (since
+version 3.0) is meant to stop the then and catch callbacks from firing. It is
+not helpful in the example of a user cancellation as described here.
```
benjamingr: Right, it does however fire `finally` callbacks which are what you want here.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+
+## Progress bar
+
+When there are asynchronous tasks that have the ability to notify progress as
+they complete it can be tempting to want that in the promise that represents
+that task. Unfortunately this is a bit of an anti-pattern. That is because the
+point of promises is to represent a value as if it was natural (like it is in
+normal synchronous code) and not to be over glorified callback management.
```
benjamingr: Also because progress is an updating event and not a single value.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+dialogs we've covered. However, there can still be some user interaction so in
+essence we are dealing with two promises.
+
+Bluebird has a way to manage more than one promise simultaneously. When you want
+to know if more then one promise completes there is a `Promise.all()` function
+that takes an array of promises and returns a new promise waiting for them all
+to resolve. But if any one is rejected the returned promise is immediately
+rejected.
+
+Bluebird also has a `Promise.race()` function which does the same thing but
```
benjamingr: Race is a terrible API we support because it's in ES2015, prefer `some` which is much more informative.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+doesn't wait for all of them to finish. That is what we want. An example how
+this might look:
+
+```javascript
+function showProgress(otherPromise) {
+  var progress = new ProgressbarDialog().show('Uploading…');
+  return Promise.race([otherPromise, promise.waitForUser()])
```
benjamingr: should be `progress.waitForUser`?

benjamingr: You also call it ProgressDialog in one place and ProgressbarDialog in another.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+and `reject` callback when you create a new promise (`new Promise(…)`) we can do
+the same patter for a progress callback.
+
+Now to the fun part. For this tutorial we will *fake* a lengthy file upload by
+using `setTimeout`. The intent is to provide a promise and to allow a progress
+to be periodically ticked away. We will expect a function to be passed which
+is called whenever the progress needs updating. And it returns a promise.
+
+```javascript
+function delayedPromise(progressCallback) {
```
benjamingr: It's important to illustrate that we're returning a promise _and_ taking a callback here - it's also important to talk about what happens if the callback throws.

docs/docs/async-dialogs.md
```diff
@@ -4,3 +4,628 @@ title: Async Dialogs
...
+}
+```
+
+When we construct our `ProgressDialog` we use the `waitForUser()` method to
+capture the user interaction promise and then use `delayedPromise()` to capture
+the fake network promise and finally `Promise.reace()` to manage the two
+simultaneously and end with a single promise as usual.
+
+```javascript
+document.addEventListener('DOMContentLoaded', function() {
```
benjamingr: again, no need to DOMContentLoaded

benjamingr:
Left some comments, thanks.

docs/docs/async-dialogs.md
spion: Race is exactly what you want here though. You wouldn't want to continue showing the progressbar if the download fails.

docs/docs/async-dialogs.md
spion: There are no non-terrible patterns when you're dealing directly with the DOM, and this one isn't particularly bad - if just `this.el` was injected in the constructor, I think it would be okay...

benjamingr: Yes, of course I mean that `this.el` needs to be injected not _all_ the selectors here. That would just be weird. Sorry for not being clearer - the problem is only with the globality.

spion: Ah okay.

docs/docs/async-dialogs.md
sukima: This is to make sure the DOM is ready. I was somewhat assuming the examples would be loaded from a `<script src=…></script>` in the `<head>`. Maybe that is a bad assumption. Perhaps that detail should be left to the student. Removing.

docs/docs/async-dialogs.md
sukima: How can I reference the handlers in the `finally()` clause? Unfortunately `removeEventListener` requires a reference to the original function passed into `addEventListener` and that is why I was forced to use the deferred pattern.

Later in the prose I explain that this is one of the few reasonable reasons for using such a pattern and follow up that in cases such as using jQuery event name spacing we can avoid this pattern and push it into the promise constructor.

I also attempted thsi with a disposer but that didn't work out right because the disposer didn't have access to the constructor's resolve/reject functions.

Any advise?

docs/docs/async-dialogs.md
sukima: This is a habit pattern that makes sure `innerHTML` is always assigned to a string and not something else accidentally. Thoughts?

docs/docs/async-dialogs.md
benjamingr: In general - it's considered a bad practice to add DOM ready events if your code is not being blindly included - you can safely include it at the end of the body section.

docs/docs/async-dialogs.md
benjamingr: Wouldn't it be better to promisify the concept of a single click event and then apply it to both buttons? It'd likely be shorter and cleaner. As for the handlers - I see your point about addEventListnener requiring a named function. I guess I'd probably nest.

docs/docs/async-dialogs.md
benjamingr: Wouldn't it be coerced to a string anyway?

docs/docs/async-dialogs.md
sukima: I see your point. I'll rework this using the button click as promise resolvers that trigger the dialog promise resolve/reject. Also I think I can avoid the problem with the handlers by attaching one click event to the dialog itself and use event delegation to tell which button is clicked.

Will experiment. Apologies for the slow responses; I just had a baby which has occupied much of my cognitive functioning.

docs/docs/async-dialogs.md
sukima: I don't believe that is the case assigning doesn't imply coercion like concatenation does. However, there is value in readability so something like `String(…)` would be more explicit.

docs/docs/async-dialogs.md
sukima: @benjamingr I can not get the nesting to work. Please help: http://jsbin.com/fucofu/edit?js,output

sukima: Oh snap I got it! Never mind:

```javascript
function promptPromise(message) {
  var dialog       = document.getElementById('dialog');
  var input        = dialog.querySelector('input');
  var okButton     = dialog.querySelector('button.ok');
  var cancelButton = dialog.querySelector('button.cancel');
  var resolver, rejector;

  dialog.querySelector('.message').innerHTML = String(message);
  dialog.className = '';

  return new Promise(function(resolve, reject) {
    dialog.addEventListener('click', function handleButtonClicks(e) {
      if (e.target.tagName !== 'BUTTON') { return true; }
      dialog.removeEventListener('click', handleButtonClicks);
      dialog.className = 'hidden';
      if (e.target === okButton) {
        resolve(input.value);
      } else if (e.target === cancelButton) {
        reject(new Error('User cancelled'));
      }
    });
  });
}
```

docs/docs/async-dialogs.md
sukima: Just curious, I've had to tell other co-workers that rule of rejecting with an Error and it's been a little bit of an up hill battle. I was able to convince with the idea that an Error can be inspected and conditional branching can be used (much like Bluebird has the `catch()` predicate) and that it mimics the standard `try {} catch {}` format.

Aside from that do you know any research or documentation that helps elaborate on this? I'd love to have better resources to fall back on the next time this anti-pattern emerges from a colleague.

benjamingr: @sukima I take it you've never had a `Unhandled rejection error: undefined` with no stack trace before :D 

Basically, not rejecting with errors makes debugging _impossible_.

benjamingr:
@sukima hey, where is this standing?

petkaantonov:
ping

sukima:
@petkaantonov @benjamingr My sincere apologies, I've been swamped lately (Just became a new dad). I was working through each comment above one by one (top down) and saving them as WIPs. Then I stopped half way through. It's going to take me a bit. Again sorry.

petkaantonov:
No worries

petkaantonov:
Doesn't seem like there are any major issues for now, I'll merge it now and the minor fixes can be done later

PR #901, opened 2015-12-03
Fix link to `new Promise` anti-pattern

jmm:
Also add the name its referred to by in `new-promise.md` to `anti-patterns.md`.

The anti-pattern content is currently duplicated in the repo / website and the wiki. Perhaps it'd be better to consolidate that content in one place or the other? (I would think in the repo / website is better.)

Here's a [squashed version](https://github.com/petkaantonov/bluebird/compare/master...jmm:new-promise-anti-patt-squashed) if you prefer.


benjamingr:
The website is correct and it should be named "explicit construction anti-pattern", I've also posted about it in StackOverflow [here](stackoverflow.com/questions/23803743/what-is-the-explicit-promise-construction-antipattern-and-how-do-i-avoid-it).

jmm:
@benjamingr Thanks for the reply. Do you mean that you'd like to change the heading & fragment ID here?: http://bluebirdjs.com/docs/anti-patterns.html#the-deferred-anti-pattern

And change `explicit constructor anti-pattern` => `explicit construction anti-pattern` here?: http://bluebirdjs.com/docs/api/new-promise.html

benjamingr:
Fixed already some time in the last month apparently

jmm:
@benjamingr
> Fixed already some time in the last month apparently

Apparently? It seems that you made the change you're referring to [here](https://github.com/petkaantonov/bluebird/commit/2637906a503fcd518b08df5ae198f8be5d8b7ce7) after the last activity on this ticket :/

The link on http://bluebirdjs.com/docs/api/new-promise.html is still broken (headline of this PR) because the fragment doesn't match, and it's now broken on http://bluebirdjs.com/docs/anti-patterns.html too.

benjamingr:
lol, I totally forgot about that :D 

Feel free to open a PR for that link and I'll merge it.

jmm:
@benjamingr Opened #940.

PR #902, opened 2015-12-04
Fixes example for replicating async.series with Promise.each

jasonmerino:
The documentation here conflicted with the documentation on [this page](http://bluebirdjs.com/docs/api/promise.each.html) as well as the actual functionality of the `Promise.each` method.


bergus:
I think instead of messing around with arrays the example should just use the new `mapSeries` method.

jasonmerino:
@bergus Changing it mapSeries is a much better idea. I've made those changes and they were much cleaner. :+1:

PR #904, opened 2015-12-07
Fix documentation for Promise.using

### LINKED ISSUE #874: Documentation issue: Promise.using
The [second paragraph](http://bluebirdjs.com/docs/api/promise.using.html) looks incomplete:

> Here is a simple example ` [has been defined] to return a proper undefined))

I would have sent a PR, but I don't know what the line is supposed to say. It's not a formatting issue.

odziem:
Some of the docs involving `Promise.using` were slightly mangled when migrating documentation from 2.x. This fixes #874.


tefla:
:+1: This is a really nice feature that is very hard to use with the documentation currently on the website.

benjamingr:
Thanks.

PR #907, opened 2015-12-09
Added monitoring functionality for promisses

lextiz:
This change introduces monitoring for pending promises. Promises are registred at creation and unregistred when their fate is sealed. At any moment library user can see all pending promises by calling Promise.getAllPendingPromises() or just the last promises in all chains by calling Promise.getLeafPendingPromises(). This basic implementation can be extended to show the bigger picture by adding bi-directional connections between promises and building DOT graph of all promises that currently exist.


src/debuggability.js
```diff
@@ -161,6 +163,46 @@ Promise.prototype._propagateFrom = function (parent, flags) {
     USE(flags);
 };
 
+Promise.monitor = function () {
+    // Pending promises are stored in array for better performance: number
+    // of simultaneously pending promises is assumed to be relatively small,
+    // thus storing promises in the map and calculating hash for each promise
+    // is more expensive then iterating trough pending promises array when
```
bergus: s/then/than/

src/debuggability.js
```diff
@@ -161,6 +163,46 @@ Promise.prototype._propagateFrom = function (parent, flags) {
     USE(flags);
 };
 
+Promise.monitor = function () {
+    // Pending promises are stored in array for better performance: number
+    // of simultaneously pending promises is assumed to be relatively small,
```
bergus: Is that really true? I could imagine `Promise.map` or similar to create a quite huge number of concurrent promises. Adding an `O(n)` step to each of them sounds desastrous.
Wouldn't a ES6 `Set` be a better fit here?

src/debuggability.js
```diff
@@ -161,6 +163,46 @@ Promise.prototype._propagateFrom = function (parent, flags) {
...
+    // thus storing promises in the map and calculating hash for each promise
+    // is more expensive then iterating trough pending promises array when
+    // each promise is finished
+    Promise.prototype._pendingPromises = [];
```
petkaantonov: Why is this array on the prototype?

bergus: I would recommend using `Promise.monitor.pendingPromises` (if you want a public property) or just a private `var`.

src/debuggability.js
petkaantonov: Cant rely on ES6,just put a unique auto-incremented id on each promise and use that in as an object key

src/debuggability.js
```diff
@@ -161,6 +163,46 @@ Promise.prototype._propagateFrom = function (parent, flags) {
...
+    function unregisterPromise() {
+        var indexOfPromise = Promise.prototype._pendingPromises.indexOf(this);
+        ASSERT(indexOfPromise >= 0);
+        Promise.prototype._pendingPromises.splice(indexOfPromise,1);
+    }
+
+    util.wrapMethodIfExistsDefineIfNot(Promise.prototype,"_promiseCreated",
+        registerPromise);
+    util.wrapMethodIfExistsDefineIfNot(Promise.prototype,"_promiseSettled",
+        unregisterPromise);
```
bergus: Why not simply replace them? Do you expect to chain multiple handlers per hook?

Also, should you guard `Promise.monitor()` against multiple invocations?

src/direct_resolve.js
```diff
@@ -9,6 +9,7 @@ function thrower() {
 
 Promise.prototype["return"] =
 Promise.prototype.thenReturn = function (value) {
+    this._setFulfilled();
```
petkaantonov: This method is equivalent to 

```js
this.then(function() {
    return value;
});
```
Why would it set `this` as fulfilled?

src/direct_resolve.js
```diff
@@ -16,11 +17,13 @@ Promise.prototype.thenReturn = function (value) {
 
 Promise.prototype["throw"] =
 Promise.prototype.thenThrow = function (reason) {
+    this._setRejected();
```
petkaantonov: Same issue as with thenReturn

src/direct_resolve.js
```diff
@@ -16,11 +17,13 @@ Promise.prototype.thenReturn = function (value) {
...
     return this._then(
         thrower, undefined, undefined, {reason: reason}, undefined);
 };
 
 Promise.prototype.catchThrow = function (reason) {
+    this._setRejected();
```
petkaantonov: Same issue as with thenReturn

src/direct_resolve.js
```diff
@@ -32,6 +35,7 @@ Promise.prototype.catchThrow = function (reason) {
 };
 
 Promise.prototype.catchReturn = function (value) {
+    this._setFulfilled();
```
petkaantonov: Same issue as with thenReturn

src/debuggability.js
```diff
@@ -161,6 +163,46 @@ Promise.prototype._propagateFrom = function (parent, flags) {
...
+
+    Promise.getAllPendingPromises = function () {
+        return Promise.prototype._pendingPromises;
```
bergus: I'd like to see a `.slice()` here, returning arrays that are later mutated is error-prone.
Or maybe don't make this a `get…` method at all, but just a public `Promise.pendingPromises` property (which can be accessed and where it's clear that the array is mutable)

src/context.js
```diff
@@ -1,10 +1,18 @@
 "use strict";
 module.exports = function(Promise) {
+
+var util = require("./util");
 var longStackTraces = false;
 var contextStack = [];
 
-Promise.prototype._promiseCreated = function() {};
-Promise.prototype._pushContext = function() {};
+util.wrapMethodIfExistsDefineIfNot(Promise.prototype, "_promiseCreated",
```
petkaantonov: what if cancellation is enabled, then monitor is enabled, then cancellation is disabled? Then monitor would not work

docs/docs/api/promise.config.md
```diff
@@ -26,16 +26,18 @@ Promise.config({
     // Enable long stack traces.
     longStackTraces: true,
     // Enable cancellation.
-    cancellation: true
+    cancellation: true,
+    // Enable promises monitoring.
+    monitor: true
```
petkaantonov: I don't think the Promise.config code was modifeid to allow this

Also monitor should be allowed to be disabled with monitor: false

docs/docs/api/environment-variables.md
```diff
@@ -20,10 +20,11 @@ Environment variables supported by 2.x:
 
 Environment variables supported by 3.x:
 
-- `BLUEBIRD_DEBUG` - If set this will enable long stack traces and warnings, unless those are explicitly disabled. Setting this to exactly `0` can be used to override `NODE_ENV=development` enabling long stack traces and warnings.
+- `BLUEBIRD_DEBUG` - If set this will enable long stack traces, warnings and promises monitoring, unless those are explicitly disabled. Setting this to exactly `0` can be used to override `NODE_ENV=development` enabling long stack traces and warnings.
```
petkaantonov: Monitoring should only be allowed to be enabled explicitly on its own.

src/debuggability.js
```diff
@@ -161,6 +163,46 @@ Promise.prototype._propagateFrom = function (parent, flags) {
...
+        ASSERT(indexOfPromise >= 0);
+        Promise.prototype._pendingPromises.splice(indexOfPromise,1);
+    }
+
+    util.wrapMethodIfExistsDefineIfNot(Promise.prototype,"_promiseCreated",
+        registerPromise);
+    util.wrapMethodIfExistsDefineIfNot(Promise.prototype,"_promiseSettled",
+        unregisterPromise);
+
+    Promise.getAllPendingPromises = function () {
```
petkaantonov: If there are other promise monitoring implementations we should have as much as the same api as they have already

lextiz:
I am working on corrections. I will answer on all the comments after the commit with corrections will be ready.

docs/docs/api/environment-variables.md
lextiz: done

src/debuggability.js
lextiz: I am not aware of similar functionality in other js promises libraries. Functionality called "monitor" in some libraries is parallel to bluebird "timeout" (e.g. https://github.com/medikoo/deferred#monitoring-unresolved-promises). Are you aware of some existing implementations?

src/direct_resolve.js
lextiz: Mistake. Corrected.

src/direct_resolve.js
lextiz: Corrected

src/direct_resolve.js
lextiz: Corrected

src/direct_resolve.js
lextiz: Corrected

src/context.js
lextiz: Did you mean longStackTraces instead of cancellation? I could not find places in code where cancellation overrides any hooks. Could you point to it? Correction commit will contain new implementation of hooks wrapper that supports both "hooking" and "unhooking", so this problem will be handled.

lextiz:
This commit is not final, I still have to fix some test failures.

src/debuggability.js
lextiz: done

src/debuggability.js
lextiz: done

src/debuggability.js
lextiz: done

src/debuggability.js
lextiz: Monitoring is the first use case that needs to place second handler on a hook. Added multiple invocations guard.

src/debuggability.js
lextiz: Not relevant for new implementation.

docs/docs/api/promise.config.md
lextiz: done

lextiz:
Fixed everything

src/debuggability.js
```diff
@@ -178,6 +188,79 @@ Promise.prototype._propagateFrom = function (parent, flags) {
...
+        Promise.monitor._pendingPromises = {};
+        Promise.monitor._promiseIdCounter = 0;
+
+        Promise.disableMonitoring = function () {
+            // No reason to clean up the id's from pending promises
+            util.unhookFrom(Promise.prototype, "_promiseCreated",
+                registerPromise);
+            util.unhookFrom(Promise.prototype,
+                "_promiseSettled", unregisterPromise);
+            delete Promise.monitor;
```
bergus: Is `delete` still performance-sensitive? Usually overwriting by `null` is enough.

src/debuggability.js
```diff
@@ -178,6 +188,79 @@ Promise.prototype._propagateFrom = function (parent, flags) {
...
+        };
+
+        util.hookTo(Promise.prototype, "_promiseCreated", registerPromise);
+        util.hookTo(Promise.prototype, "_promiseSettled", unregisterPromise);
+
+        Promise.getPendingPromises = function () {
+            var result = [];
+            // Object.values() comes only in EC7
```
bergus: Do you mean E*S*7?

src/util.js
```diff
@@ -276,6 +276,51 @@ function copyDescriptors(from, to, filter) {
...
+    } else if (typeof existingMethodImpl === "undefined") {
+        prototypeObject[methodName] = function () {
+            var extensions = prototypeObject[methodName].extensions;
+            for (var key in extensions) {
+                if (extensions.hasOwnProperty(key)) {
+                    extensions[key].apply(this, arguments);
+                }
+            }
+        };
+        // Using object instead of array of functions to allow optimizations
```
bergus: I don't understand. Which optimisations are you talking about?

src/util.js
```diff
@@ -276,6 +276,51 @@ function copyDescriptors(from, to, filter) {
...
+}
+
+function unhookFrom(prototypeObject, methodName, extension) {
+    var extensions = prototypeObject[methodName].extensions;
+    for (var i = 0; i < extensions; i++) {
+        if (extensions[i] === extension) {
+            extensions.splice(i, 1);
+        }
+    }
+}
```
bergus: The loop is non-functional. Also you really want to use an array instead of an object, otherwise there's no `splice` method.

src/util.js
```diff
@@ -276,6 +276,51 @@ function copyDescriptors(from, to, filter) {
...
+// This function should not be used for methods that return a value
+// Using this function implies small performance overhead, should be used only
+// for debug methods
+function hookTo(prototypeObject, methodName, extension) {
+    var existingMethodImpl = prototypeObject[methodName];
+    if (typeof existingMethodImpl === "function" &&
+        existingMethodImpl.extensions) {
+        var extensions = existingMethodImpl.extensions;
+        extensions[extension.toString()] = extension;
+        // Performance optimization: remove noops
```
bergus: Why would you add noops to this in the first place? I'd recommend either passing `null` or a unique `function noop(){}` as the `extension`, and just test for that by value.

src/util.js
```diff
@@ -276,6 +276,51 @@ function copyDescriptors(from, to, filter) {
...
+        for (var key in extensions) {
+            if (extensions.hasOwnProperty(key) &&
+                extensions[key].toString().replace(/\s/g, "") ===
+                "function(){}") {
```
bergus: Really, use a shared `noop` function (like the `(Function.prototype)` function) and compare it by identity.
This code does e.g. not work in Firefox which returns `function () {\n"use strict";\n}` for empty strict mode functions.

bergus:
What are all these file mode changes in the `src` directory about?

src/debuggability.js
lextiz: Is performance critical in function that disables monitoring? I think it is cleaner to remove the whole property. Also I want to delete all the inner objects as soon as possible.

src/debuggability.js
lextiz: Yes. Corrected.

src/util.js
lextiz: As far as I understand saving functions on object property and not in array allows JS execution engine to optimize out useless frame in call stack (or another part of call stack). If it would be saved in array longStackTraces test would fail with message "Expected 2 jumps but saw 3 jumps".

src/util.js
lextiz: Fixed. Added unit test for this.

src/util.js
lextiz: noops are needed for situation when debugging capabilities are disabled, in this case hook functions could even not be packaged

src/debuggability.js
petkaantonov: It makes `Promise` a hash map which makes reading all properties from `Promise` much slower. Like `Promise.all`

lextiz:
Some cygwin-git issues, will fix the file permissions.

src/debuggability.js
```diff
@@ -178,6 +188,79 @@ Promise.prototype._propagateFrom = function (parent, flags) {
     USE(flags);
 };
 
+Promise.enableMonitoring = function () {
+    function registerPromise() {
+        if (Promise.monitor) {
+            Promise.monitor._promiseIdCounter++;
+            if (Promise.monitor._promiseIdCounter === Number.MAX_VALUE) {
```
petkaantonov: This check should be removed, there is no way this is never happening:

Firstly, because an int cannot be incremented with `++` after  `9007199254740992`:

```js
var a = 9007199254740992;
var l = 1000; while(l--) a++`
// 9007199254740992
a;
```

Secondly, if you create a **million** promises per **second**, it would take **285 years** to reach **9007199254740992**.

So in practice these conditions will never be executed and they are therefore dead code.

src/util.js
lextiz: Yes, using function.toString() was not a good idea, also because of cases like "function() {[native code]}". Will replace with static counter.

src/util.js
bergus: I understand that you need the noop hook-methods, but I don't see any reason to place empty functions in these `extension` collections.
Just call `util.hookTo(Promise.prototype, "_promise…", null);` instead of `util.hookTo(Promise.prototype, "_promise…", function() {});` in the setup above, and initialise the `methodImpl` but ignore non-function values when adding to `extensions`.

src/debuggability.js
lextiz: done

src/debuggability.js
lextiz: done

src/util.js
lextiz: done

src/util.js
```diff
@@ -276,6 +276,45 @@ function copyDescriptors(from, to, filter) {
...
+            for (var key in extensions) {
+                if (extensions.hasOwnProperty(key)) {
+                    extensions[key].apply(this, arguments);
+                }
+            }
+        };
+        // Using object instead of array of functions to allow optimizations
+        prototypeObject[methodName].extensions = {};
+        if (typeof extension === "function") {
+            extension._hookExtensionId = ++_hookIdCounter;
```
lextiz: Maybe @petkaantonov could explain what exactly is optimized out in stack trace in this case?

src/util.js
lextiz: Maybe @petkaantonov could explain what exactly is optimized out in stack trace in this case?

lextiz:
Done with fixes.

src/util.js
bergus: OK now I'm curious :-)

docs/docs/api/promise.monitor.md
```diff
@@ -0,0 +1,38 @@
...
+title: Promise.monitor
+---
+
+
+[← Back To API Reference](/docs/api-reference.html)
+<div class="api-code-section"><markdown>
+##Promise.monitor
+
+```js
+Promise.enableMonitoring() -> undefined
```
petkaantonov: Why is this needed if `Promise.config({monitor: true})` can be used?

lextiz: It is done to be compliant to standard in code: longStackTraces can be enabled via envariable, config, and method call, this feature is similar, no?

lextiz:
I have prepared one more PR: limiting promise chain depth, it uses hookTo and unhookFrom methods. Shall I wait until this PR is merged and submit the next feature in separate PR or flush it here?

docs/docs/api/promise.monitor.md
petkaantonov: `Promise.longStackTraces()` is only for backwards compatibility with 2.x

lextiz: done

docs/docs/api/promise.config.md
```diff
@@ -26,16 +27,18 @@ Promise.config({
...
 });
 ```
 
 <hr>
 
-In Node.js you may configure warnings and long stack traces for the entire process using environment variables:
+In Node.js you may configure warnings, long stack traces and promises monitoring for the entire process using environment variables:
```
petkaantonov: promises monitoring -> promise monitoring

src/debuggability.js
```diff
@@ -128,6 +130,68 @@ Promise.hasLongStackTraces = function () {
...
+        Promise.monitor._promiseIdCounter++;
+        this._promiseId = Promise.monitor._promiseIdCounter;
+        Promise.monitor._pendingPromises[Promise.monitor._promiseIdCounter]
+            = this;
+    }
+}
+
+function unregisterPromise() {
+    if (Promise.monitor)
+        delete Promise.monitor._pendingPromises[this._promiseId];
```
petkaantonov: This probably  doesn't need to be accessible from the promise object? It could just be a private closure variable

src/debuggability.js
```diff
@@ -128,6 +130,68 @@ Promise.hasLongStackTraces = function () {
...
+        Promise.monitor._pendingPromises = {};
+        Promise.monitor._promiseIdCounter = 0;
+
+        util.hookTo(Promise.prototype, "_promiseCreated", registerPromise);
+        util.hookTo(Promise.prototype, "_promiseSettled", unregisterPromise);
+
+        Promise.getPendingPromises = function () {
+            var result = [];
+            // Object.values() comes only in ES7
+            for (var key in Promise.monitor._pendingPromises) {
```
petkaantonov: Use Object keys from the "es5.js" module

src/debuggability.js
```diff
@@ -128,6 +130,68 @@ Promise.hasLongStackTraces = function () {
...
+    if (Promise.monitor)
+        delete Promise.monitor._pendingPromises[this._promiseId];
+}
+
+function enableMonitoring () {
+    if (!Promise.monitor) {
+        // Property that holds monitoring related info,
+        // existence of it means that monitoring feature is currently enabled
+        Promise.monitor = {};
+        Promise.monitor._pendingPromises = {};
```
petkaantonov: These again don't need to be public

src/util.js
```diff
@@ -276,6 +276,45 @@ function copyDescriptors(from, to, filter) {
...
+        if (!existingMethodImpl.extensions) {
+            throw new Error("Trying to extend overriden method,"+
+                " please use util.js:hookTo() to extend it");
+        }
+        extension._hookExtensionId = ++_hookIdCounter;
+        existingMethodImpl.extensions[extension._hookExtensionId] = extension;
+    } else if (typeof existingMethodImpl === "undefined") {
+        prototypeObject[methodName] = function () {
+            var extensions = prototypeObject[methodName].extensions;
+            for (var key in extensions) {
```
petkaantonov: Use Object.keys from the es5.js module

src/util.js
```diff
@@ -276,6 +276,45 @@ function copyDescriptors(from, to, filter) {
     }
 }
 
+var _hookIdCounter = 0;
+// This function should not be used for methods that return a value
+// Using this function implies small performance overhead, should be used only
```
petkaantonov: I am not sure if reflecting property keys and looping through them for every method call is acceptable even for debugging mode. I need to think of a better way to do this.

docs/docs/api/promise.config.md
lextiz: done, changed all the occurrences

src/debuggability.js
lextiz: On which closure should I put it?

src/debuggability.js
lextiz: done

src/util.js
lextiz: done

src/util.js
lextiz: I assume you are saying this in terms of performance. If so, I will write a short benchmark, comparing "function wrapping" method and the current one. If the current method is significantly worse, than I can use "function wrapping" and rebuild the hook handler function on each "unhook".

src/debuggability.js
lextiz: Same here. I do not understand on which closure shall I put this property to be accessible from all the places that use it.

src/util.js
lextiz: You are right, the benchmark shows, that "implementations loop" method is much worse than "function wrapping", will change it. Attaching benchmark code, FYI, it can be run as is browser console.

```javascript
var objectKeys = null;
if (this === undefined) {
    objectKeys = Object.keys;
} else {
    var has = {}.hasOwnProperty;
	objectKeys = function (o) {
        var ret = [];
        for (var key in o) {
            if (has.call(o, key)) {
                ret.push(key);
            }
        }
        return ret;
    };
}

function hookToArrayImpl(prototypeObject, methodName, extension) {
    var existingMethodImpl = prototypeObject[methodName];
    if (typeof existingMethodImpl === "function" &&
        typeof extension === "function") {
        if (!existingMethodImpl.extensions) {
            throw new Error("Trying to extend overriden method,"+
                " please use util.js:hookTo() to extend it");
        }
        extension._hookExtensionId = ++_hookIdCounter;
        existingMethodImpl.extensions[extension._hookExtensionId] = extension;
    } else if (typeof existingMethodImpl === "undefined") {
        prototypeObject[methodName] = function () {
            var extensions = prototypeObject[methodName].extensions;
            var keys = objectKeys(extensions);
            for (var i = 0; i < keys.length; i++) {
                extensions[keys[i]].apply(this, arguments);
            }
        };
        // Using object instead of array of functions to allow optimizations
        prototypeObject[methodName].extensions = {};
        if (typeof extension === "function") {
            extension._hookExtensionId = ++_hookIdCounter;
            prototypeObject[methodName].extensions[_hookIdCounter] = extension;
        }
    } else {
        throw new Error("Trying to wrap " + typeof existingMethodImpl +
            ", expecting a function or undefined");
    }
}

function hookToWrapImpl(prototypeObject, methodName, extension) {
    var existingMethodImpl = prototypeObject[methodName];
    if (typeof existingMethodImpl === "function") {
        prototypeObject[methodName] = function () {
            existingMethodImpl.apply(this, arguments);
            extension.apply(this, arguments);
        };
    } else if (typeof existingMethodImpl === "undefined") {
        prototypeObject[methodName] = extension;
    } else {
        throw new Error("Trying to wrap " + typeof existingMethodImpl + 
            ", expecting a function or undefined");
    }
}

function emptyFunction () {};
function smallFunction () { window.a = Math.random() };
function mediumFunction () { window.a = Math.random() * Math.random()};
function largeFunction () { for (var i=0; i<100; i++) {window.a += Math.random()}};


function runBenchmark(hookToImpl, extensions) {
    var Class = function() {};
	for (var i=0; i<extensions.length; i++) {
	    hookToImpl.call(null, Class.prototype, "method", extensions[i]);
    }
	var object = new Class();
	var start = new Date();
    for (var i=0; i<100000; i++) {
        object.method();
    }
	return  new Date() - start;
}

// Most common use case
console.log("hookToArrayImpl, standard use case: " + runBenchmark(hookToArrayImpl, [emptyFunction,smallFunction]));
console.log("hookToWrapImpl, standard use case: " + runBenchmark(hookToWrapImpl, [emptyFunction,smallFunction]));

console.log("hookToArrayImpl, heavy use case: " + runBenchmark(hookToArrayImpl, [emptyFunction,smallFunction,mediumFunction,largeFunction]));
console.log("hookToWrapImpl, heavy use case: " + runBenchmark(hookToWrapImpl, [emptyFunction,smallFunction,mediumFunction,largeFunction]));
```

src/debuggability.js
petkaantonov: in this file somewhere, like the other closure variables. Or is it used from other files?

src/util.js
lextiz: OK, this problem was solved with current implementation, I have checked its performance with same benchmark, it is same like manually wrapping functions (but manual wrapping does not have support for unhook).

src/debuggability.js
lextiz: It is not used from another files. I will move it out from monitor property. I have tried to put everything related to monitoring functionality on this property to have better encapsulation, when the monitoring is disabled everything could be deleted.

src/debuggability.js
lextiz: done

lextiz:
Done with fixes.

lextiz:
I have finished with fixes. Do you have additional remarks? When is this PR planned to be merged?

petkaantonov:
Have you rebased it with the latest?

lextiz:
Rebased.

lextiz:
Moved to this PR: https://github.com/petkaantonov/bluebird/pull/922

lextiz:
https://github.com/petkaantonov/bluebird/pull/922 is the same but the diff is more clean, it does not include rebase related diffirences.

BridgeAR:
@lextiz you do not have to open a new pull request to have a clean diff. What you can do is to rebase with git. So what you did so far is:

You
1. fork the project
2. clone your fork on your device
3. edit something
4. push the change to your origin
5. open a pull request
6. pull changes from uptream/master into your pull request that require a merge

But if you want to change something in that pull request, especially to resolve conflicts you can do the following (rebase):

```
git remote add upstream https://github.com/petkaantonov/bluebird.git
git fetch upstream
git rebase -i upstream/master
# resolve all conflicts
git push -f
```
Or you could also use `git pull --rebase`

That way you can change the history of your pull request and the diff is going to be clean again.

lextiz:
I did the rebase. And than I did pull from my fork, to be able to push there (this was probably wrong step, will use "pull --rebase" next time). Then I pushed. I could probably fix everything in this PR, but it would have lots of unnecessary commits. This is the reason I have opened a new PR, it seems cleaner to me.

BridgeAR:
If you rebase you won't have any unnecessary commits and you're able to use `squash` or `fixup` to merge different commits into a single new one. If you then use `push -f` you manipulate the commit history and all old commits will be gone and only the new ones exist.

PR #914, opened 2015-12-15
Update promise.promisify.md to fix argument name

yotamofek:
Fixed name of wrong argument in documentation


benjamingr:
The argument name was correct, just unorthodox. I guess this is better, thanks.

PR #922, opened 2015-12-17
Promise lifecycle events subscriptions and extended debugging/monitoring capabilities (Monitoring for pending promises)

lextiz:
Promises are registred at creation and unregistred when their fate is sealed. At any moment library user can see all pending promises by calling Promise.getPendingPromises() or just the last promises in all chains by calling Promise.getLeafPendingPromises().

PR #929, opened 2015-12-21
Promise lifecycle event subscriptions

lextiz:
https://github.com/petkaantonov/bluebird/pull/907
https://github.com/petkaantonov/bluebird/pull/922

### DIFF
--- a/docs/docs/features.md
+++ b/docs/docs/features.md
@@ -13,6 +13,7 @@ title: Features
 - [Resource management](#resource-management)
 - [Cancellation and timeouts](#cancellation-and-timeouts)
 - [Scoped prototypes](#scoped-prototypes)
+- [Promises monitoring](#promises-monitoring)
 - [Async/Await](#async-await)
 
 ##Synchronous inspection
@@ -232,10 +233,7 @@ See [installation](install.html) on how to enable long stack traces in your envi
 
 ###Error pattern matching
 
-Perhaps the greatest thing about promises is that it unifies all error handling into one mechanism where errors propagate automatically and have to be explicitly ignored. However
-
-
-
+Perhaps the greatest thing about promises is that it unifies all error handling into one mechanism where errors propagate automatically and have to be explicitly ignored.
 
 ###Warnings
 
@@ -243,7 +241,36 @@ Promises can have a steep learning curve and it doesn't help that promise standa
 
 See [installation](install.html) on how to enable warnings in your environment.
 
+###Promise monitoring
 
+This feature enables subscription to promise lifecycle events via standard global events mechanisms in browsers and Node.js. Following events are available: "promiseCreated", "promiseChained", "promiseFulfilled", "promiseRejected", "promiseResolved" and "promiseCancelled".
+This feature has to be explicitly enabled, see: [Promise.config](api/promise.config.html).
+Subscription to promise lifecycle events is possible in these ways:
+1. In Node.js via process.on() method:
+```
+process.on("promiseCreated", function(eventName, promise) {
+    // eventName - "promiseCreated"
+    // promise - promise object that was created
+})
+```
+2. In browsers via self.addEventListener() method:
+```
+self.addEventListener("promiseChained", function(event) {
+    // event.type - "promiseCreated"
+    // event.details.promise - promise object that was chained to child promise
+    // event.details.child - child promise object
+})
+```
+3. In browsers via overriding handler method on self:
+Note: In this method event names have "on" prefix and have to be in lower case.
+This method implies single event handler function per event type.
+```
+self.onpromisechained = function(eventName, promise, child) {
+    // eventName - "promiseChained"
+    // promise - promise object that was chained to child promise
+    // child - child promise object
+})
+```
 
 ##Resource management
 
@@ -265,5 +292,4 @@ Your library can then use `var Promise = require("bluebird-extended");` and do w
 
 
 ##Async/Await
-
 
--- a/src/debuggability.js
+++ b/src/debuggability.js
@@ -141,6 +141,84 @@ Promise.hasLongStackTraces = function () {
     return config.longStackTraces && longStackTracesIsSupported();
 };
 
+function createFireDomEvent() {
+    try {
+        var event = document.createEvent("CustomEvent");
+        event.initCustomEvent("testingtheevent", false, true, {});
+        global.dispatchEvent(event);
+        return function(name, event) {
+            var domEvent = document.createEvent("CustomEvent");
+            domEvent.initCustomEvent(name.toLowerCase(), false, true, event);
+            return !global.dispatchEvent(domEvent);
+        };
+    } catch (e) {}
+}
+
+var fireDomEvent = createFireDomEvent();
+var fireGlobalEvent = (function() {
+    if (util.isNode) {
+        return function() {
+            return process.emit.apply(process, arguments);
+        };
+    } else {
+        if (!global) {
+            return function() {
+                return false;
+            };
+        }
+        return function(name) {
+            var methodName = "on" + name.toLowerCase();
+            var method = global[methodName];
+            if (!method) return false;
+            method.apply(global, [].slice.call(arguments, 1));
+            return true;
+        };
+    }
+})();
+
+function generatePromiseLifecycleEventObject(name, promise) {
+    return {promise: promise};
+}
+
+var eventToObjectGenerator = {
+    promiseCreated: generatePromiseLifecycleEventObject,
+    promiseFulfilled: generatePromiseLifecycleEventObject,
+    promiseRejected: generatePromiseLifecycleEventObject,
+    promiseResolved: generatePromiseLifecycleEventObject,
+    promiseCancelled: generatePromiseLifecycleEventObject,
+    promiseChained: function(name, promise, child) {
+        return {promise: promise, child: child};
+    },
+    warning: function(name, warning) {
+        return {warning: warning};
+    },
+    unhandledRejection: function (name, reason, promise) {
+        return {reason: reason, promise: promise};
+    },
+    rejectionHandled: generatePromiseLifecycleEventObject
+};
+
+var fireEventImpl = function () {
+    var name = arguments[0];
+    var eventFired = false;
+    try {
+        eventFired = fireGlobalEvent.apply(null, arguments);
+    } catch (e) {
+        async.throwLater(e);
+        eventFired = true;
+    }
+    if (fireDomEvent) {
+        try {
+            eventFired = eventFired || fireDomEvent(name,
+                    eventToObjectGenerator[name].apply(null, arguments));
+        } catch (e) {
+            async.throwLater(e);
+            eventFired = true;
+        }
+    }
+    return eventFired;
+};
+
 Promise.config = function(opts) {
     opts = Object(opts);
     if ("longStackTraces" in opts) {
@@ -176,6 +254,20 @@ Promise.config = function(opts) {
         Promise.prototype._execute = cancellationExecute;
         propagateFromFunction = cancellationPropagateFrom;
         config.cancellation = true;
+    }
+    if ("monitoring" in opts) {
+        if (opts.monitoring && !config.monitoring) {
+            if (opts.monitoring === "no_dom_events") {
+                fireDomEvent = undefined;
+            } else {
+                fireDomEvent = createFireDomEvent();
+            }
+            Promise.fireEvent = fireEventImpl;
+            config.monitoring = opts.monitoring;
+        } else if (config.monitoring) {
+            Promise.fireEvent = function (){};
+            config.monitoring = false;
+        }
     }
 };
 
@@ -336,6 +428,7 @@ function warn(message, shouldUseOwnTrace, promise) {
         warning.stack = parsed.message + "\n" + parsed.stack.join("\n");
     }
     formatAndLogError(warning, "", true);
+    Promise.fireEvent("warning", warning);
 }
 
 function reconstructStack(message, stacks) {
@@ -466,30 +559,12 @@ function fireRejectionEvent(name, localHandler, reason, promise) {
         async.throwLater(e);
     }
 
-    var globalEventFired = false;
-    try {
-        globalEventFired = fireGlobalEvent(name, reason, promise);
-    } catch (e) {
-        globalEventFired = true;
-        async.throwLater(e);
-    }
-
-    var domEventFired = false;
-    if (fireDomEvent) {
-        try {
-            domEventFired = fireDomEvent(name.toLowerCase(), {
-                reason: reason,
-                promise: promise
-            });
-        } catch (e) {
-            domEventFired = true;
-            async.throwLater(e);
+    if (name === UNHANDLED_REJECTION_EVENT) {
+        if (!fireEventImpl(name, reason, promise) && !localEventFired) {
+            formatAndLogError(reason, UNHANDLED_REJECTION_HEADER);
         }
-    }
-
-    if (!globalEventFired && !localEventFired && !domEventFired &&
-        name === UNHANDLED_REJECTION_EVENT) {
-        formatAndLogError(reason, UNHANDLED_REJECTION_HEADER);
+    } else {
+        fireEventImpl(name, promise);
     }
 }
 
@@ -753,59 +828,6 @@ var captureStackTrace = (function stackDetection() {
 
 })([]);
 
-var fireDomEvent;
-var fireGlobalEvent = (function() {
-    if (util.isNode) {
-        return function(name, reason, promise) {
-            if (name === REJECTION_HANDLED_EVENT) {
-                return process.emit(name, promise);
-            } else {
-                return process.emit(name, reason, promise);
-            }
-        };
-    } else {
-        var globalObject = typeof self !== "undefined" ? self :
-                     typeof window !== "undefined" ? window :
-                     typeof global !== "undefined" ? global :
-                     this !== undefined ? this : null;
-
-        if (!globalObject) {
-            return function() {
-                return false;
-            };
-        }
-
-        try {
-            var event = document.createEvent("CustomEvent");
-            event.initCustomEvent("testingtheevent", false, true, {});
-            globalObject.dispatchEvent(event);
-            fireDomEvent = function(type, detail) {
-                var event = document.createEvent("CustomEvent");
-                event.initCustomEvent(type, false, true, detail);
-                return !globalObject.dispatchEvent(event);
-            };
-        } catch (e) {}
-
-        var toWindowMethodNameMap = {};
-        toWindowMethodNameMap[UNHANDLED_REJECTION_EVENT] = ("on" +
-            UNHANDLED_REJECTION_EVENT).toLowerCase();
-        toWindowMethodNameMap[REJECTION_HANDLED_EVENT] = ("on" +
-            REJECTION_HANDLED_EVENT).toLowerCase();
-
-        return function(name, reason, promise) {
-            var methodName = toWindowMethodNameMap[name];
-            var method = globalObject[methodName];
-            if (!method) return false;
-            if (name === REJECTION_HANDLED_EVENT) {
-                method.call(globalObject, promise);
-            } else {
-                method.call(globalObject, reason, promise);
-            }
-            return true;
-        };
-    }
-})();
-
 if (typeof console !== "undefined" && typeof console.warn !== "undefined") {
     printWarning = function (message) {
         console.warn(message);
@@ -826,7 +848,8 @@ if (typeof console !== "undefined" && typeof console.warn !== "undefined") {
 var config = {
     warnings: warnings,
     longStackTraces: false,
-    cancellation: false
+    cancellation: false,
+    monitoring: false
 };
 
 if (longStackTraces) Promise.longStackTraces();
@@ -841,6 +864,9 @@ return {
     cancellation: function() {
         return config.cancellation;
     },
+    monitoring: function() {
+        return config.monitoring;
+    },
     propagateFromFunction: function() {
         return propagateFromFunction;
     },
@@ -851,6 +877,8 @@ return {
     setBounds: setBounds,
     warn: warn,
     deprecated: deprecated,
-    CapturedTrace: CapturedTrace
+    CapturedTrace: CapturedTrace,
+    fireDomEvent: fireDomEvent,
+    fireGlobalEvent: fireGlobalEvent
 };
 };
--- a/src/util.js
+++ b/src/util.js
@@ -8,6 +8,13 @@ var canEvaluate = typeof navigator == "undefined";
 //compiler, so it is isolated
 var errorObj = {e: {}};
 var tryCatchTarget;
+var isNode = typeof process !== "undefined" &&
+    classString(process).toLowerCase() === "[object process]";
+var globalObject = typeof self !== "undefined" ? self :
+    typeof window !== "undefined" ? window :
+    typeof global !== "undefined" ? global :
+    this !== undefined ? this : null;
+
 function tryCatcher() {
     try {
         var target = tryCatchTarget;
@@ -306,9 +313,6 @@ if (typeof Symbol !== "undefined" && Symbol.iterator) {
     };
 }
 
-var isNode = typeof process !== "undefined" &&
-        classString(process).toLowerCase() === "[object process]";
-
 function env(key, def) {
     return isNode ? process.env[key] : def;
 }
@@ -342,7 +346,8 @@ var ret = {
     hasDevTools: typeof chrome !== "undefined" && chrome &&
                  typeof chrome.loadTimes === "function",
     isNode: isNode,
-    env: env
+    env: env,
+    global: globalObject
 };
 ret.isRecentNode = ret.isNode && (function() {
     var version = process.versions.node.split(".").map(Number);
--- a/tools/test.js
+++ b/tools/test.js
@@ -108,7 +108,7 @@ function needsFreshProcess(testName) {
 function runTestGroup(testGroup, options, progress) {
     return jobRunner.run(mochaRunner, {
         isolated: !options.singleTest,
-        log: options.singleTest,
+        log: options.singleTest || options.verbose,
         progress: progress,
         context: {
             testGroup: testGroup,
@@ -170,7 +170,8 @@ var options = {
         ? argv["fake-timers"] : true,
     jsHint: typeof argv["js-hint"] === "boolean" ? argv["js-hint"] : true,
     nw: !!argv.nw,
-    nwPath: argv["nw-path"]
+    nwPath: argv["nw-path"],
+    verbose: !!argv.verbose
 };
 
 
@@ -246,8 +247,12 @@ var testResults = Promise.join(tests, buildResult, function(tests) {
     } else if (singleTest) {
         return runTestGroup(tests, options);
     } else {
-        process.stdout.cursorTo(0, 0);
-        process.stdout.clearScreenDown();
+        // Node.js on git bash in Windows 10 and Cygwin do not have
+        // cursorTo and clear screenDown implementations
+        if (process.stdout.cursorTo && process.stdout.clearScreenDown) {
+	        process.stdout.cursorTo(0, 0);
+            process.stdout.clearScreenDown();
+	    }
         tableLogger.addTests(tests);
         return Promise.map(combineTests(tests), function(testGroup, index) {
             return runTestGroup(testGroup, options, function(test) {
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ test/browser/bundle.js
 js/*
 zalgo.js
 coverage/*
+.idea/*
--- a/docs/docs/api/promise.config.md
+++ b/docs/docs/api/promise.config.md
@@ -13,20 +13,23 @@ title: Promise.config
 Promise.config(Object {
     warnings: boolean=false,
     longStackTraces: boolean=false,
-    cancellation: boolean=false
+    cancellation: boolean=false,
+    monitoring: boolean=false
 } options) -> undefined;
 ```
 
-Configure long stack traces, warnings and cancellation. Note that even though `false` is the default here, a development environment might be detected which automatically enables long stack traces and warnings.
+Configure long stack traces, warnings, monitoring and cancellation. Note that even though `false` is the default here, a development environment might be detected which automatically enables long stack traces and warnings.
 
 ```js
 Promise.config({
-    // Enable warnings.
+    // Enable warnings
     warnings: true,
-    // Enable long stack traces.
+    // Enable long stack traces
     longStackTraces: true,
-    // Enable cancellation.
-    cancellation: true
+    // Enable cancellation
+    cancellation: true,
+    // Enable monitoring
+    monitoring: true
 });
 ```
 
--- a/package.json
+++ b/package.json
@@ -49,11 +49,13 @@
     "co": "^4.2.0",
     "cross-spawn": "^0.2.3",
     "glob": "^4.3.2",
+    "grunt": "^0.4.1",
     "grunt-saucelabs": "~8.4.1",
     "highland": "^2.3.0",
     "istanbul": "^0.3.5",
     "jshint": "^2.6.0",
     "jshint-stylish": "~0.2.0",
+    "kefir": "^2.4.1",
     "mkdirp": "~0.5.0",
     "mocha": "~2.1",
     "open": "~0.0.5",
@@ -62,8 +64,7 @@
     "rx": "^2.3.25",
     "serve-static": "^1.7.1",
     "sinon": "~1.7.3",
-    "uglify-js": "~2.4.16",
-    "kefir": "^2.4.1"
+    "uglify-js": "~2.4.16"
   },
   "readmeFilename": "README.md",
   "main": "./js/release/bluebird.js",
--- a/src/promise.js
+++ b/src/promise.js
@@ -44,13 +44,14 @@ var INTERNAL = function(){};
 var APPLY = {};
 var NEXT_FILTER = {};
 var tryConvertToPromise = require("./thenables")(Promise, INTERNAL);
-var PromiseArray =
-    require("./promise_array")(Promise, INTERNAL,
-                               tryConvertToPromise, apiRejection, Proxyable);
 var Context = require("./context")(Promise);
  /*jshint unused:false*/
 var createContext = Context.create;
+Promise.fireEvent = function(){};
 var debug = require("./debuggability")(Promise, Context);
+var PromiseArray =
+    require("./promise_array")(Promise, INTERNAL,
+    tryConvertToPromise, apiRejection, Proxyable);
 var CapturedTrace = debug.CapturedTrace;
 var PassThroughHandlerContext =
     require("./finally")(Promise, tryConvertToPromise);
@@ -78,6 +79,7 @@ function Promise(executor) {
         this._resolveFromExecutor(executor);
     }
     this._promiseCreated();
+    Promise.fireEvent("promiseCreated", this);
 }
 
 Promise.prototype.toString = function () {
@@ -238,6 +240,7 @@ Promise.prototype._then = function (
                 receiver = target === this ? undefined : this._boundTo;
             }
         }
+        Promise.fireEvent("promiseChained", this, promise);
     }
 
     var domain = getDomain();
@@ -291,14 +294,17 @@ Promise.prototype._setLength = function (len) {
 
 Promise.prototype._setFulfilled = function () {
     this._bitField = this._bitField | IS_FULFILLED;
+    Promise.fireEvent("promiseFulfilled", this);
 };
 
 Promise.prototype._setRejected = function () {
     this._bitField = this._bitField | IS_REJECTED;
+    Promise.fireEvent("promiseRejected", this);
 };
 
 Promise.prototype._setFollowing = function () {
     this._bitField = this._bitField | IS_FOLLOWING;
+    Promise.fireEvent("promiseResolved", this);
 };
 
 Promise.prototype._setIsFinal = function () {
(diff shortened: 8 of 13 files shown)


lextiz:
Ready for review

docs/docs/api/promise.on.md
```diff
@@ -0,0 +1,54 @@
...
+title: Promise.on
+---
+
+
+[← Back To API Reference](/docs/api-reference.html)
+<div class="api-code-section"><markdown>
+##Promise.on
+
+```js
+Promise.on(String eventName("created"|"chained"|"fulfilled"|"rejected"|"following"|"cancelled"), Function handler, [Boolean simpleEventAPI]) -> undefined
```
petkaantonov: "following" is internal terminology, it should probably be "resolved". Events should probably be on `process` or `window` so that they can be tracked using one set of listeners for all bluebird instances.

lextiz: following->resolved is done. Regarding bluebird instances: each application probably uses single bluebird instance, different instances come from 3rd party dependencies, right? Why would somebody like to subscribe to these events that happen outside of his own code?
If you want it on (window|process) anyway, should it be something like (window|process).Promise.on()?

spion: @lextiz If you're using other promise-supporting libraries that use Bluebird, you will most likely derive promises from them, e.g.

```js
function myFunctionality(blah) {
  return dbLibrary.queryReturningPromise(query).then(doSomeProcessing);
}
```

Originally Bluebird used non-global events, but a bunch of issues were quickly opened that stemmed from the above, so as a result it now has [global events](http://bluebirdjs.com/docs/api/error-management-configuration.html#global-rejection-events)

lextiz: @spion Exactly this is what I mean: why would you be interested when some promise is created in dbLibrary? I think in most cases this is not important information, unless you want to debug the library you are using. Also it makes this events not very consistent: lets say dbLibrary uses bluebird and dbLibraryUtilityLibrary uses Q then the Q lifecycle events will not be reported. 
@petkaantonov For performance reasons I have to attach event handlers to functions inside single bluebird instance. This means that to maintain events on global level I would need a list of all bluebird instances that exist, is this currently possible to get? Would it be acceptable I add  an array with bluebird instances to (process|window)?

spion: @lextiz but that promise is created by library's bluebird instance. Therefore, the promise derived with `then` will also be created by that instance, and as such will not be tracked. You would have to use Promise.resolve(db.query(...)).then(...) instead of simply chaining `then`, and you'd have to do it every time you use an external library.

lextiz: @spion Ok, now I understand the problem. Global events subscriptions are safer to implement than handling this problem in other ways. What do you think is this acceptable to store an array with bluebird instances on (process|window)?

lextiz: Done.

lextiz:
Ready for review.

benjamingr:
IIRC - the reason we had `onPossiblyUnhandledRejection` as a function is that you can't make a typo. I wonder if that wouldn't be a better pattern than `on` - to just have `Promise.onFulfilled(..)` or even `Promise.on.fulfilled(...)` that way if you call a function/hook that doesn't exist you get a typeerror immediately .

lextiz:
@petkaantonov wanted the same API as in RSVP. I can add this additional API as syntactic sugar if you guys think it is convenient.

lextiz:
Kind code review reminder...

docs/docs/features.md
```diff
@@ -13,6 +13,7 @@ title: Features
 - [Resource management](#resource-management)
 - [Cancellation and timeouts](#cancellation-and-timeouts)
 - [Scoped prototypes](#scoped-prototypes)
+- [Promises monitoring](#promises-monitoring)
```
petkaantonov: Perhaps link should be added to the API table of contents as well

src/debuggability.js
```diff
@@ -141,6 +141,69 @@ Promise.hasLongStackTraces = function () {
     return config.longStackTraces && longStackTracesIsSupported();
 };
 
+var supportedEvents = ["created", "chained", "fulfilled",
+    "rejected", "resolved", "cancelled"];
+
+Promise.on = function (eventName, hookFunction, simpleEventApi) {
+    if (supportedEvents.indexOf(eventName)<0) {
```
petkaantonov: Please use white space  consistent  with the project:

`if (supportedEvents.indexOf(eventName) < 0) {`

src/debuggability.js
```diff
@@ -141,6 +141,69 @@ Promise.hasLongStackTraces = function () {
...
+        hookFunction._eventHandler = newHook;
+    }
+    var instances = util.global.__BLUEBIRD_INSTANCES__;
+    for (var i = 0; i < instances.length; i++) {
+        util.hookTo(instances[i].prototype, "_hook_" + eventName, newHook);
+    }
+};
+
+Promise.off = function (eventName, hookFunction) {
+    if (supportedEvents.indexOf(eventName)<0) {
```
petkaantonov: Whitespace

src/util.js
```diff
@@ -276,6 +281,86 @@ function copyDescriptors(from, to, filter) {
...
+    if (!prototypeObject[methodName].extensions) {
+        throw new Error("Trying to unhook from method, that was not extended");
+    }
+    var index = prototypeObject[methodName].extensions.indexOf(extension);
+    if (index < 0) {
+        throw new Error("Trying to unhook extension function " + extension +
+            " that was not hooked to current prototype, extensions that were" +
+            " hooked are: " + prototypeObject[methodName].extensions);
+    }
+    prototypeObject[methodName].extensions.splice(index,1);
```
petkaantonov: whitespace: `index, 1`

src/util.js
```diff
@@ -276,6 +281,86 @@ function copyDescriptors(from, to, filter) {
...
+                var prev = previousImpl;
+                var next = extensions[i];
+                return function () {
+                    prev.apply(this, arguments);
+                    next.apply(this, arguments);
+                };
+            }());
+        }
+    }
+    currentImpl.extensions = extensions;
```
petkaantonov: Is there a way to do this without adding new properties to functions? (Which changes their hidden class and opts out of function optimizations for .call and .apply for instance)

src/debuggability.js
```diff
@@ -141,6 +141,69 @@ Promise.hasLongStackTraces = function () {
     return config.longStackTraces && longStackTracesIsSupported();
 };
 
+var supportedEvents = ["created", "chained", "fulfilled",
+    "rejected", "resolved", "cancelled"];
+
+Promise.on = function (eventName, hookFunction, simpleEventApi) {
```
petkaantonov: Just one API is sufficient. It will make this easier to use and understand and it should reduce some internal code as well I hope (currently the hooking stuff is very bizarre and complex).

src/debuggability.js
```diff
@@ -141,6 +141,69 @@ Promise.hasLongStackTraces = function () {
...
+                detail: this.isRejected() ? this.reason :
+                    this.isFulfilled() ? this.value : null,
+                //event.label //is this relevant?
+                timeStamp: Date.now(),
+                stack: this._trace ? this._trace : null
+            });
+        };
+        hookFunction._eventHandler = newHook;
+    }
+    var instances = util.global.__BLUEBIRD_INSTANCES__;
```
petkaantonov: So when `Promise.on` is called, that is same as calling it on all promise libraries? Doesn't make sense.

src/debuggability.js
lextiz: RSVP API alone does not allow access to promise object, which is necessary for some debugging capabilities, it does not allow to put _promiseId property on Promise for example.

petkaantonov: You can add the promises as additional properties of the event object, that is fully compatible

lextiz: Agree. But what about performance: creating the event object and extra frame for each event will cause some overhead. Obligating all users of this API to such overhead is OK with you?

docs/docs/features.md
lextiz: done

src/debuggability.js
lextiz: done

src/util.js
lextiz: done

src/debuggability.js
lextiz: done

src/debuggability.js
lextiz: done

src/util.js
lextiz: done

docs/docs/api/promise.off.md
```diff
@@ -0,0 +1,31 @@
...
+
+
+[← Back To API Reference](/docs/api-reference.html)
+<div class="api-code-section"><markdown>
+##Promise.off
+
+```js
+Promise.off(String eventName("created"|"chained"|"fulfilled"|"rejected"|"following"|"cancelled"), Function handler) -> undefined
+```
+Removes a handler form promise lifecycle event (see: [on](promise.on.html).
```
benjamingr: `form -> from a`?

docs/docs/api/promise.off.md
```diff
@@ -0,0 +1,31 @@
...
+Given event name and handler function that were used for subscription, removes the subscription.
```
benjamingr: What about something similar to what jQuery has in the docs? `The .off() method removes event handlers that were attached with .on().` and shortly explain the params?

docs/docs/api/promise.on.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+[← Back To API Reference](/docs/api-reference.html)
+<div class="api-code-section"><markdown>
+##Promise.on
+
+```js
+Promise.on(String eventName("created"|"chained"|"fulfilled"|"rejected"|"resolved"|"cancelled"), Function handler, [Boolean simpleEventAPI]) -> undefined
+```
+Subscribes a handler function for an event from promise lifecycle ("created", "chained", "fulfilled", "rejected", "resolved" and "cancelled").
```
benjamingr: We need to explain when each event is fired - even in one line. We also need to emphasize whether or not events happen across bluebird instances.

docs/docs/api/promise.on.md
```diff
@@ -0,0 +1,54 @@
...
+Event handlers can be of 2 kinds:
+1) Handler receives an event object in given format:
+```js
+function(Object:{guid: Number, childGuid: Number, eventName: String, detail: any, timeStamp: Date, stack: Trace}) -> undefined;
```
benjamingr: I think it would be more readable to explain each of these parameters in a separate line. That signature is long and what `stack` means in the context of a `resolved` event isn't very clear here.

docs/docs/api/promise.off.md
lextiz: done

docs/docs/api/promise.off.md
lextiz: done

docs/docs/api/promise.on.md
lextiz: done

docs/docs/api/promise.on.md
lextiz: done

src/bluebird.js
```diff
@@ -7,5 +7,14 @@ function noConflict() {
     return bluebird;
 }
 var bluebird = require("./promise")();
+var util = require("./util");
 bluebird.noConflict = noConflict;
 module.exports = bluebird;
+
+// Bluebird versions older then 3.1.1 (including) do not
+// support promise lifecycle events
+if (util.global.__BLUEBIRD_INSTANCES__) {
```
benjamingr: Hmm, I'm not sure this would work everywhere - I recall a few environments (react native, nw.js, firefox extensions) where global detection might fail. 

For the very least this should be in a try/catch. I didn't even know we exported a util.global - that's new I guess.

src/context.js
```diff
@@ -1,10 +1,17 @@
 "use strict";
 module.exports = function(Promise) {
+
+var util = require("./util");
 var longStackTraces = false;
 var contextStack = [];
 
-Promise.prototype._promiseCreated = function() {};
-Promise.prototype._pushContext = function() {};
+util.hookTo(Promise.prototype, "_hook_created", null);
```
benjamingr: Given all these changes - can you please run the benchkmarks before/after this PR and with montioring turned on and off and report the results? 

We need to be extra careful that users with monitoring turned off do not have to pay for it.

src/debuggability.js
```diff
@@ -141,6 +141,63 @@ Promise.hasLongStackTraces = function () {
     return config.longStackTraces && longStackTracesIsSupported();
 };
 
+var supportedEvents = ["created", "chained", "fulfilled",
+    "rejected", "resolved", "cancelled"];
+
+Promise.on = function (eventName, hookFunction, simpleEventApi) {
+    if (supportedEvents.indexOf(eventName) < 0) {
+        throw new Error("You can only subscribe to these events:" +
```
benjamingr: Error -> TypeError

src/debuggability.js
```diff
@@ -141,6 +141,63 @@ Promise.hasLongStackTraces = function () {
...
+    } else {
+        newHook = function (other) {
+            hookFunction({
+                guid: this._promiseId ? this._promiseId : null,
+                childGuid: (other && other._promiseId) ?
+                    other._promiseId : null,
+                eventName: eventName,
+                detail: this.isRejected() ? this.reason :
+                    this.isFulfilled() ? this.value : null,
+                //event.label //is this relevant?
```
benjamingr: @petkaantonov ^ question

src/debuggability.js
```diff
@@ -141,6 +141,63 @@ Promise.hasLongStackTraces = function () {
...
+            });
+        };
+        hookFunction._eventHandler = newHook;
+    }
+    util.hookTo(Promise.prototype, "_hook_" + eventName, newHook);
+};
+
+Promise.off = function (eventName, hookFunction) {
+    if (supportedEvents.indexOf(eventName) < 0) {
+        throw new Error("You can only subscribe to these events:"
```
benjamingr: Error -> TypeError

src/debuggability.js
```diff
@@ -141,6 +141,63 @@ Promise.hasLongStackTraces = function () {
...
+            + supportedEvents);
+    }
+    util.unhookFrom(Promise.prototype, "_hook_" + eventName,
+        hookFunction._eventHandler ? hookFunction._eventHandler :
+            hookFunction);
+};
+
+// Event handlers can be placed on (window|process).__PROMISE_INSTRUMENTATION__
```
benjamingr: Any particular reason that if they're placed on `window/process` they're not using the window/process event handlers like `unhandledRejection` but using their own `on/off`?

src/debuggability.js
```diff
@@ -141,6 +141,63 @@ Promise.hasLongStackTraces = function () {
...
+// object
+if (util.global && typeof util.global.__PROMISE_INSTRUMENTATION__ ===
+    "object") {
+    var callbacks = util.global.__PROMISE_INSTRUMENTATION__;
+    for (var eventName in callbacks) {
```
benjamingr: Any particular reason we support passing an object in? Is this documented?

src/debuggability.js
```diff
@@ -764,12 +821,7 @@ var fireGlobalEvent = (function() {
             }
         };
     } else {
-        var globalObject = typeof self !== "undefined" ? self :
```
benjamingr: I wonder if this can cause TDZ errors if someone makes a `self` block scoped variable in node for example. Although it doesn't sound like a really problematic case since in practice I suspect this will be a non-issue.

src/util.js
```diff
@@ -276,6 +281,85 @@ function copyDescriptors(from, to, filter) {
...
+    var existingMethodImpl = prototypeObject[methodName];
+    var extensions = null;
+    if (typeof existingMethodImpl === "function") {
+        extensions = global.eventsToHandlers[methodName];
+        if (!extensions) {
+            extensions = global.eventsToHandlers[methodName]
+                = [];
+        }
+        if (extensions.indexOf(extension) >= 0) {
+            throw new Error("Trying to register same extension method twice");
```
benjamingr: `TypeError`

src/util.js
```diff
@@ -276,6 +281,85 @@ function copyDescriptors(from, to, filter) {
...
+// This function should not be used for methods that return a value
+// Using this function implies small performance overhead, should be used only
+// for debug methods
+function hookTo(prototypeObject, methodName, extension) {
+    var existingMethodImpl = prototypeObject[methodName];
+    var extensions = null;
+    if (typeof existingMethodImpl === "function") {
+        extensions = global.eventsToHandlers[methodName];
+        if (!extensions) {
+            extensions = global.eventsToHandlers[methodName]
```
benjamingr: Hmm, same question from above - why do we need our own event implementation rather than fire events like in the global hooks? Did we just decide to go with local instrumentation? If so that's fine, if not we should be consistent in event handling.

src/util.js
```diff
@@ -276,6 +281,85 @@ function copyDescriptors(from, to, filter) {
...
+            extensions = [extension];
+            prototypeObject[methodName] = generateFunctionFromExtensions(
+                extensions, methodName);
+        } else {
+            extensions = [];
+            prototypeObject[methodName] =
+                generateFunctionFromExtensions(extensions, methodName);
+        }
+    } else {
+        throw new Error("Trying to wrap " + typeof existingMethodImpl +
```
benjamingr: TypeError

src/util.js
```diff
@@ -276,6 +281,85 @@ function copyDescriptors(from, to, filter) {
...
+            ", expecting a function or undefined");
+    }
+}
+
+function unhookFrom(prototypeObject, methodName, extension) {
+    var extensions = global.eventsToHandlers[methodName];
+    if (!Array.isArray(extensions)) {
```
benjamingr: `Array.isArray` would not work in old IE, or probably even in some evergreen browsers. `extensions instanceof Array` would probably work.

src/util.js
```diff
@@ -276,6 +281,85 @@ function copyDescriptors(from, to, filter) {
...
+        throw new Error("Trying to unhook from method, that was not extended");
```
benjamingr: TypeError

src/util.js
```diff
@@ -276,6 +281,85 @@ function copyDescriptors(from, to, filter) {
...
+    if (index < 0) {
+        throw new Error("Trying to unhook extension function " + extension +
+            " that was not hooked to current prototype, extensions that were" +
+            " hooked are: " + extensions);
+    }
+    extensions.splice(index, 1);
+    prototypeObject[methodName] = generateFunctionFromExtensions(extensions);
+}
+
+function generateFunctionFromExtensions(extensions, methodName) {
```
benjamingr: This method looks kind of needlessly complicated and not very performant. Any reason why we do all this wrapping?

tools/test.js
```diff
@@ -246,8 +246,12 @@ var testResults = Promise.join(tests, buildResult, function(tests) {
     } else if (singleTest) {
         return runTestGroup(tests, options);
     } else {
-        process.stdout.cursorTo(0, 0);
-        process.stdout.clearScreenDown();
+        // Node.js on git bash in Windows 10 and Cygwin do not have
```
benjamingr: Oh god thank you :D

benjamingr:
LGTM with comments. Only really critical one is using `Array.isArray`, Good work.

src/bluebird.js
lextiz: Removed __BLUEBIRD_INSTANCES__, it is not needed anymore after refactoring.

benjamingr: Thanks, let me know when the changes are ready for review.

lextiz: Of course I will, it will take a while.

src/util.js
lextiz: Why TypeError?

src/debuggability.js
lextiz: done

src/debuggability.js
lextiz: done

src/util.js
benjamingr: Because it's an error with the type, it's kind of a stretch though. What error type would you recommend? I guess since this is internal it's not a big deal.

lextiz: Not a big deal, I just ask to understand if there is something special about TypeError. I would recommend "Error". Done.

src/debuggability.js
lextiz: I think firing global events is more complex in terms of maintainability and performance.
Regarding \__PROMISE_INSTRUMENTATION\__: as far as I understand @petkaantonov wanted the events to be fired same in same way as in RSVP, here https://github.com/tildeio/rsvp.js/blob/fcd59034548508a2126b9a3cd1c7132288a7b04f/lib/rsvp.js#L39 I have found sngle reference to window in context of events in RSVP code. From RSVP code I understand that library users are expected to put the event handlers on window.\__PROMISE_INSTRUMENTATION\__ properties, this is the way in RSVP I did it the same. Of course there is an option to fire DOM events (performance?) or to call window.onSomethingHappened() methods (name compatibility), as it is done in debuggability.js:fireGlobalEvent() but these methods look less convenient to me.

src/debuggability.js
lextiz: If you mean __PROMISE_INSTRUMENTATION__ object than I will document it as soon as we decide to keep going with this (__PROMISE_INSTRUMENTATION__ on window) option (see previous comment).

src/debuggability.js
lextiz: Agree. There is no reason to worry at all, because the definition is in global scope of utils.js file, library users that write "self = this" will not affect this. The only case when user overrides self object in global scope should not be handled in my opinion.

src/util.js
lextiz: Two reasons to go with local instrumentation:
1) There are name compatibility issues
2) Users would have to handle multiple event handlers themselves (all the functionality of hookTo unhookFrom functions in utils.js) and in this case unsubscribing from the event would be a problem for them.

src/util.js
lextiz: done

src/util.js
lextiz: done

src/util.js
lextiz: This method is not meant to be very performant it is called only from on() and off(). In first implementation I have stored the hooks in array and called them sequentially, but this prevented from V8 to apply some optimizations and performance suffered. This is the reason why I am using nested handlers calls, that each handler calls the next.

src/context.js
lextiz: There is no sense to turn monitoring feature on and off because it does not add self executed code, it only adds methods that could be invoked, impacting the library size only. It could be made separate optional module, like "call_get".
Performance measurement results (doxbee-sequential):

__before the change:__
promises-bluebird-generator.js                  250       29.55
promises-bluebird.js                            359       45.42
__after the change:__
promises-bluebird-generator.js                  228       29.32
promises-bluebird.js                            360       45.05

I see that the variance of the benchmark is much higher than between 2 measurements, so we can conclude that there is no performance degradation.

lextiz:
Just to summarize: all code review remarks were fixed. Some questions are still open and, I guess, are waiting for @petkaantonov to have a look:
1) Should simple event handler API be left (:pray:) or removed (despite its performance efficiency comparing to second method)?
2) Local instrumentation (:+1:) vs global events (:-1:).
3) Is event.label field relevant, if yes what should be inside?

petkaantonov:
1. No multiple apis
2. Global events on `process` or `window`
3. Not needed

lextiz:
Regarding 2: on() and off() APIs allow registration of multiple handlers for single event. How multiple handlers would be possible if "onpromisecreated" like functions on window are used as event handlers?

petkaantonov:
in node `process.on()` allows multiple listeners

so does `self.addEventListener` in browsers

lextiz:
Ok, now I get it. Are on() and off() methods needed in bluebird then? User could register to events using global APIs.

petkaantonov:
> Ok, now I get it. Are on() and off() methods needed in bluebird then? 

No they are not needed. See how rejection events are done, the Promise.onPossiblyUnhandledRejection etc are only for legacy reasons and backward compatibility

lextiz:
If global events handling is used, for new events to be consistent with existing events (rejection) following changes have to be done to fireGlobalEvent code:
1) Event handler API should be changed to uniform one (from (name, promise) or (name, reason, promise) arguments to (name, event) arguments, where event would hold all relevant info, including reason and promise)
2) If local events are not wanted onPossiblyUnhandledRejection has to be deprecated, otherwise on() and off() methods are needed.
These changes would break the API. Also, as far as I understand, using DOM events is less performant than local events (also @domenic does not like dispatchEvent for some reason: https://github.com/tildeio/rsvp.js/issues/7). Should I go ahead with these changes? I think the possible options are:
1) Use local events and leave on() and off() methods, as I was asked to asked to do earlier (to be compliant to RSVP API)
2) Unify event handlers API as described above
3) Implement lifecycle events aside from rejection events, if no consistent and uniform API is needed
What do you think?

lextiz:
There are de-facto 2 APIs for global events in browser (at least for onerror):
1) What I call "simple" API: all relevant objects are passed to event handler as arguments (like onerror(errorMsg, url, lineNumber) or onunhandledrejection(reason, promise) or possibly onpromisechained(promise, child)). This kind of event subscriptions does not support multiple events, but they are performant.
2) Standard DOM events: event object is passed to event handler. Less performant option.
First API is performant and second allows multiple events. Existing local instrumentation code (when used with "simple" API) solves these 2 problems of existing browser APIs: it does allow multiple events and it is more performant than HTML5 DOM events.

petkaantonov:
What do you mean by "consistent"?

lextiz:
Consistent examples:
1) node:                        process.emit("unhandledRejection", eventObject)
    new browser events: window.dispatchEvent(eventObject);
    old browser events:   window.onunhandledrejection.call(window, eventObject);
2) node:                        process.emit("unhandledRejection", reason, promise*)
    new browser events: not implemented
    old browser events:   window.onunhandledrejection.call(window, reason, promise);
Not consistent example:
1) node:                        process.emit("unhandledRejection", reason, promise)
    new browser events: window.dispatchEvent(eventObject);
    old browser events:   window.onunhandledrejection.call(window, reason, promise);
This situation is not consistent because:
1) Some event handlers get events in one format (event object) and some in another (event attributes as arguments)
2) Performance of one method differs from another
3) If library user wants to choose between modern DOM events and old "handler on window" api he has to understand the performance implications of first method and that second method does not handle multiple event handlers. If the decision is done implicitly in code this might cause misunderstanding.

* reason, promise are event properties, represented as arguments list in case of unhandledRejection event
* eventObject is object containing the event name and other relevant event properties (reason and promise in case of unhandledRejection event)

petkaantonov:
Do you mean that node and browser events are not "consistent" with each other? That is by design, they even have different event names `"unhandledRejection"` vs `"unhandledrejection"`

benjamingr:
That is also the name of the events in node and the browser for native promises. So while we picked them we don't control them any more anyway. 

The names were chosen to be consistent with how each platform defines its event names.

lextiz:
Yes I mean that old browser "handler on window" apis (and possibly node also) are not consistent with new browser events APIs (dispatchEvent). This inconsistency is not related to bluebird, but in my opinion this should not prevent from bluebird to supply uniform API.

The issue is not about the names, it is about arguments of event handler:
How would you expect me to implement "onPromiseChained" for example?
In node environment:
process.emit("promiseChained", promise, child) __or__ process.emit("promiseChained", eventObject)?
In browser environment:
window.dispatchEvent(eventObject). This has no alternatives and introduces API weakness.
and if window.onpromisechained function exists:
window.onunhandledrejection(promise, child) __or__ window.onunhandledrejection(eventObject)?

benjamingr:
I think we should consider scoping the events if we fire them globally, i.e. 'bluebird:promiseChained' etc.

events have to be global because there are lots of scenarios where you have more than one copy of bluebird (for example - if you use Sequelize or RethinkDB)

petkaantonov:
We will not have "consistent" or "uniform" api with node. Dom events has event objects with properties, node has multiple arguments.

lextiz:
@benjamingr Events are stored globally already in current implementation, with on() and off(). Regarding events names: as far as I understand you are obligated to use same event names as native promises, no?
@petkaantonov The implementation will become much simpler in this case, I will rewrite it. Anyway I think that such implementation is not enough intuitive and exposes user to the problems described above.

lextiz:
Done with fixes, but benchmark looks bad. The code that was removed (local instrumentation) worked with almost 0 performance overhead, as far as I understand 0 overhead is not possible with global events.

petkaantonov:
Why not? Why would events being global affect it?

petkaantonov:
These events should not be fired unless monitoring is enabled, when monitoring is disabled those functions should be noops hence no performance overhead from the feature when it's disabled

lextiz:
There is an objective reason why global events are much slower: we do not have information if user has subscribed for certain event and each time event is fired we have to call process.emit (or browsers analogues). process.emit will probably lookup for handlers in map and invoke them sequentially, optimizations of such flow by V8 (and others) are limited.

On the other hand local instrumentation receives the handler functions at the moment of subscription and generates one handler function from all handlers, this handler function is invoked straight after event happened and allows V8 to perform optimizations, this is the reason for good performance.

Why don't you like local instrumentation? The events are stored globally in previous implementation and multiple instances of bluebird are handled.

lextiz:
Regarding disabling the monitoring feature: every feature could be disabled, but in the end person who uses this feature will experience slowness. Local and global handling could coexist side by side as it is done in onPossiblyUnhandledRejection.

petkaantonov:
Are you saying someone enables monitoring and then doesn't attach handlers, so the global events will be fired "for nothing"?

lextiz:
Mentioned use case is less interesting in my opinion. I mean that users who subscribe to global events will have big performance overhead. If on() and off() methods would exist, user would still have same functionality with zero performance overhead for using the events.

petkaantonov:
If someone has enabled monitoring and therefore already opted out of performance, why would the tiny or non-existent performance difference between global and local events matter?

lextiz:
Benchmark shows that that the difference is approximately 2 times slower. I have thought to use this feature in production, because it could have zero overhead. Why monitoring feature has to cause performance overhead? In RSVP for example it does not.

benjamingr:
@lextiz you can check if an event on `process` has handlers attached to it by looking at the return value of `emitter.listenerCount`. 

In browsers the performance difference should probably not matter anyway. 

It would be interesting to see tooling emerge based on the monitoring feature :) (cc @paulirish @domenic )

petkaantonov:
It doesn't matter because if monitoring is enabled, presumable there are handlers attached which actually do something that will completely hide the overhead from dispatchEvent or process.emit.

lextiz:
In standard use case user would attach only few handlers, not for all events.

petkaantonov:
Optimizing for the case where the user has enabled monitoring but doesn't listen to all events is insane

lextiz:
There is an example of perfromant usage of these event APIs: the extended debugging capabilities (https://github.com/petkaantonov/bluebird/pull/922, RIP). If each capability is enabled and disabled separately the code in handlers does almost nothing (e.g. attaching id to promise). In this case the overhead of process.emit and dispatchEvent is considerable. Anyway heavy code in handlers for promise lifecycle events should be rare, no? I think at least we should allow a possibility to use the events in performant way.

petkaantonov:
```js
var now = process.hrtime();
for (var i = 0; i < 1000000; ++i) {
    process.emit("unlistenedEvent");
}
var diff = process.hrtime(now);
console.log((diff[0] * 1000) + (diff[1] / 1e6), "ms");
```

`94.805573 'ms'`

Meanwhile other far more crucial debugging features like long stack traces  easily have 100x more overhead. It's not even a discussion.

benjamingr:
How about we do this:

 - In any case if it is disabled there should be absolutely zero performance penalty.
 - If it is enabled and you're in node - you can check the `listenerCount` for what you want.
 - If it is enabled and you're in the browser - fire the event. 

I'm open to ideas in the browser - but I'd like to see more detailed analysis first showing that it causes the said slowdown.

petkaantonov:
For the price of `listenerCount()` you could have called `emit()` already.

lextiz:
Fare enough. I will profile this and share the results (or code fix).

lextiz:
I have executed a performance test:
```javascript
var bluebird = require('./js/release/bluebird.js');

function emptyHandler(promise) {}

function regularHandlerCreated(promise) {
    if (!promise.hasOwnProperty("_id")) promise["_id"] = 1;
    else promise._id++;
}

function regularHandlerChained(promise, child) {
    promise["next"] = child;
    child["prev"] = promise;
}

function heavyHandler(promise) {
    var q;
    for (var i = 0; i < 100; i++) {
        q = Math.random();
    }
}

function test(handler, handler2) {
    //bluebird.on("created", handler, true);
    //bluebird.on("chained", handler2 ? handler2 : handler, true);
    process.on("promiseCreated", handler);
    process.on("promiseChained", handler2 ? handler2 : handler);
    var a;
    var start = Date.now();
    for (var i = 0; i < 1000000; i++) {
        a = new bluebird(function () {
        });
        a = a.then(function () {
        });
        a = a.then(function () {
        });
    }
    console.log(Date.now()-start);
    //bluebird.off("created", handler);
    //bluebird.off("chained", handler2 ? handler2 : handler);
    process.removeAllListeners("promiseCreated");
    process.removeAllListeners("promiseChained");
}

//test(emptyHandler);
test(regularHandlerCreated, regularHandlerChained);
//test(heavyHandler);
```

The results are (in milliseconds):
__Current implementation:__
Monitoring disabled: 139
Emit non-existing event: 904
Emit existing event with empty handler: 395
Emit existing event with regular handler (e.g.: assign id to a promise): 715
Emit existing event with heavy handler (e.g.: 100 random numbers): 3802

__Previous implementation:__
Monitoring disabled: 139
Emit non-existing event: not relevant
Emit existing event with empty handler: 215
Emit existing event with regular handler: 465
Emit existing event with heavy handler: 3612

I have also profiled the code, did not find bottlenecks in code of both implementations. Profiling results are given for "regular handlers", that are corresponding to most common use case. 
Call | Self | Self % | Total | Total %
__Current implementation:__
*hasOwnProperty native v8natives.js:111:30 8 17.0% 11 23.4%
*emit events.js:117:44 5 10.6% 20 42.6%
*Promise._then promise.js:215:36 4 8.5% 23 48.9%
__Previous implementation:__
~<anonymous> perftest.js:34:35 2 12.5% 2 12.5%
~runInThisContext node.js:882:28 2 12.5% 2 12.5%
FastNewClosureStub 2 12.5% 2 12.5%

The conclusions are:
1. My previous guess that process.emit significantly slows down the monitoring is right
2. In most common use case current implementation slows down bluebird x5.14, previous implementation x3.35.

lextiz:
If you stall want to proceed with global events my suggestion is:
1. Add switch for monitoring mode
2. Add switch for browsers that would allow to suppress global DOM events and invoke only handlers on window (for people who do not want the overhead, they could use handlers on window with no performance overhead)
Is this acceptable? Should I proceed with fix?

petkaantonov:
> Add switch for monitoring mode

There *is* supposed to be a switch for monitoring mode, that is also never implicitly enabled

lextiz:
Done with fixes, ready for review.

lextiz:
Benchmark results with disabled monitoring are exactly the same as in master branch.

petkaantonov:
Thanks I will look at it when I get time :)

tyscorp:
@lextiz thanks for all of your work on this! It looks really useful.

lextiz:
You are welcome.

lextiz:
Kind reminder...

petkaantonov:
Sorry for delay, ill merge this in this weekend

lextiz:
Thanks =)

petkaantonov:
In Chrome 48 there are 18 tests failing. Didn't check any other platform

lextiz:
I have fixed the typo which was causing browser tests failures.
After the rebase I have run into interesting issue: the test "github-689" fails, but only when I run all the tests together. The tests "github-689" (regress.js) and "cancels the promise chain within a domain GH963" (cancel.js) look like they are not intended to run in parallel: one test uses process.domain and second overrides it, in the middle of the test the overridden value (process.domain) becomes undefined. I can't see any logical connection between my change and the test failure, except race condition timing change. @petkaantonov Could you please advice me: are tests "github-689" and "GH963" intended to run in parallel?

lextiz:
I have made the test github-689 that overrides the process.doman to override only process.domain.bind() instead if it exists, this should make the test agnostic to concurrent usage of domain.

I have run the tests on latest Chrome, Firefox, Edge and IE on Windows 10 and also on sauce labs, the results are exactly the same on master and in my branch.

benjamingr:
@lextiz if you'd like we have a browserstack account you could use for testing IE9 and such. Thanks a lot for the work by the way it's just a large amount of code so we're extra careful about it.

lextiz:
בכיף. Great, how could I use your account? Sauce labs seems to be in a bad mood =(.

benjamingr:
Can you send me your email (to benji @ tipranks . com)?

lextiz:
I have checked all popular platform/browser combinations (all browsers on win7 and win 10, safari on latest mac, android(s6)) on browserstack: tests pass everywhere, except some (probably known) failures on IE11 and edge at both master and my branch. The change looks ready...

lextiz:
https://github.com/petkaantonov/bluebird/issues/980.

petkaantonov:
Can you squash your changes into a single commit

lextiz:
done

petkaantonov:
Landed in 3.2.0

lextiz:
Thanks for qualitative code review guys!

petkaantonov:
Reverted due to https://github.com/petkaantonov/bluebird/issues/988

lextiz:
I have validated https://github.com/petkaantonov/bluebird/pull/989 fixes https://github.com/petkaantonov/bluebird/pull/988. What happens to this change now?

petkaantonov:
It will be relanded once we have verified it works again. I had to revert because the crashing was critical and didn't have time to deeply investigate it

lextiz:
All right, tell me if some action is needed from my side. The bug is that: "global" was meant to be "util.global", just global was not defined, I wonder how tests did not fail in browsers with this bug.

Issue #927, opened 2015-12-18
Promise.coroutine should respect the arity of the generator function being wrapped?


renegare:
Hi guys,

My use case is in the context of a cucumber-js step definition:

```
  this.Then(/^I should do (.+)$/, Promise.coroutine(function* (something) {
    expect(yield this.do(something)).to.be.true
  }))
```

Currently the above step when executed will return the following error from cucumber-js:

```
Error: step definition has 0 arguments, should have 1 (if synchronous or returning a promise) or 2 (if accepting a callback)
```

I appreciate this could be something cucumber-js should avoid throwing an error on or I solve it locally in my project, but I feel it would be a useful change within bluebird for all.

Thoughts?

phpnode:
This is not something that can be fixed by bluebird at all. It's cucumber-js which is making a pretty silly decision to rely on `Function.prototype.length`

renegare:
Fair enough and I agree more or less. cucumberjs aside, Do you mind explaining why is this something that cannot or should not be implemented?

I don't see this as bug, more of a missing feature.

(P.S: Currently looking into how to programmatically create anonymous wrapper functions with matching arity)

petkaantonov:
The only way to do it is with `new Function` or creating a function for every possible arity. Since in browsers `new Function` cannot be used due to CSP, that leaves creating a function for every possible arity which is obviously unacceptable.

renegare:
Fair enough, that does make it unreasonable.

phpnode:
@petkaantonov in node, Firefox and Chrome (I couldn't immediately check other environments) it is actually possible to do this using `Object.defineProperty()`:
```
> function a (a, b, c) {}
undefined
> a.length
3
> Object.defineProperty(a, 'length', {value: 1111})
[Function: a]
> a.length
1111
> 
```

petkaantonov:
@phpnode oh that's cool

renegare:
great news! this is "a way" to achieve what I was looking for:

```
var Promise = require('bluebird')

module.exports = function(gen) {
  var arity = (new Array(gen.length)).fill('a').map((i,n) => i + n)
  eval('function func (' + arity.join(', ')+ ') { return Promise.coroutine(gen).apply(this, arguments) }')
  return func
}
```

@phpnode not entirely clear on how your example works, but if theres anything I can do, let me know.

petkaantonov:
[...]

PR #933, opened 2015-12-24
Promise.coroutine should respect the arity of the generator

renegare:
#927

### DIFF
--- a/src/generators.js
+++ b/src/generators.js
@@ -12,6 +12,7 @@ var util = require("./util");
 var errorObj = util.errorObj;
 var tryCatch = util.tryCatch;
 var yieldHandlers = [];
+var defineProperty = require("./es5").defineProperty;
 
 function promiseFromYieldHandler(value, yieldHandlers, traceParent) {
     for (var i = 0; i < yieldHandlers.length; ++i) {
@@ -178,7 +179,8 @@ Promise.coroutine = function (generatorFunction, options) {
     var yieldHandler = Object(options).yieldHandler;
     var PromiseSpawn$ = PromiseSpawn;
     var stack = new Error().stack;
-    return function () {
+
+    function wrap () {
         var generator = generatorFunction.apply(this, arguments);
         var spawn = new PromiseSpawn$(undefined, undefined, yieldHandler,
                                       stack);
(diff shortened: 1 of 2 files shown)


benjamingr:
What's the behavior of this on old IE?

petkaantonov:
on IE arity will be 0

benjamingr:
@renegare why the mocha test changes? Also, can you please add some tests to this?

spion:
@benjamingr looks like a test has already been added. Are there any browser obstacles to merging this?

renegare:
@all happy new year and all that jazz ... been awol but back.

@benjamingr ... old IE ... do they even support generators?

@petkaantonov  could you suggest a better place to put tests? i presummed ```test/mocha/generator.js``` is for testing the usage of generators. no?

@spion I have not taken a look, I'm on osx. Is this a manual process to verify in IE (and all other browsers) or is there a more simpler way to do so?

Apologies on the slowness of getting back. If you can give me clearer direction or someone would like take this over, feel free :)

benjamingr:
> old IE ... do they even support generators?

Only through regenerator and similar tools. I was just concerned this would not break things there. Testing it - it looks fine. 

> .. could you suggest a better place to put tests? i presummed test/mocha/generator.js is for testing the usage of generators. no?

I would add two/three tests for different scenarios (function with lots of arguments, function without arguments, etc). 

Happy new year.

renegare:
Cool fair enough, I'll update the tests and run through some manual tests and report back.

as for regenerator, do you have a link to this? sounds interesting ... I've only recently come to appreciate gens #noO0b

renegare:
found it: https://github.com/facebook/regenerator

is it me or do we refuse to let old IEs die ...?

benjamingr:
@renegare we build things so people use them - if people still use older versions of IE, we have nothing to do about it except if we're willing to lose money for it.

renegare:
@benjamingr I'm not being paid to do this ...;) but hey its just a thought ...

... the next es7,8,9 more polyfills to keep old IEs on board ...

... then again you may be right about money. It does keep me in a job.

petkaantonov:
Merged in 3.1.2

renegare:
Thanks guys.

PR #934, opened 2015-12-24
Update error.md

lionvs:
Fix console.lag to console.log


benjamingr:
Read the text about it - that's intentional. The surrounding text literally gives it as an example of an error :)

lionvs:
Yeah, I've read it later, sorry about that:)

PR #935, opened 2015-12-28
Foreign thenables with internal data keeps resolved value.

chuym:
This patch makes foreign thenables, with an internal data representation, keep their resolved values when passing through methods like `Promise.all`

Example of thenable object with internal data representation:

    var thenable = {
        _v: 2,
        then: function (f) {
            this._v = f(v);
            return this;
        }
    };

such thenables allow for chained `then` methods that keep transformations from previously called `then` methods, however, if this thenable is passed to `Promise.all` the resolved value is lost because the resolver function from `src/thenables.js` does not return the resolved value again.

### DIFF
--- a/src/thenables.js
+++ b/src/thenables.js
@@ -71,12 +71,14 @@ function doThenable(x, then, context) {
         if (!promise) return;
         promise._resolveCallback(value);
         promise = null;
+        return value;
     }
 
     function reject(reason) {
         if (!promise) return;
         promise._rejectCallback(reason, synchronous, true);
         promise = null;
+        return reason;
     }
     return ret;
 }
--- a/test/mocha/promise_array.js
+++ b/test/mocha/promise_array.js
@@ -175,4 +175,18 @@ describe("Promise.all-test", function () {
         });
     });
 
+    specify("should keep resolved value for foreign thenables with internal data", function() {
+        var foreign = {
+            _v: 2,
+            then: function (f) {
+                this._v = f(this._v);
+                return this._v;
+            }
+        };
+        return Promise.all([foreign, foreign]).then(
+            function (results) {
+                assert.equal(results[0], results[1]);
+            });
+    });
+
 });


petkaantonov:
I don't understand what is different from:

    then: function (f) {
        this._v = v;
        f(v);
        return this;
    }

Also, pretty sure that the functions must return undefined (@domenic)

chuym:
The example is just a quick crude snippet to represent the idea behind the PR not meant to be taken seriously, in practice `then` must return a new promise and if the resolver return a value, the returned promise should be resolved with that returned value (same for rejected as well)

`then` [must return a promise](https://promisesaplus.com/#point-40):

`onFulfiled` [can optionally return a value](https://promisesaplus.com/#point-41), when that happens, the promise returned by the `then` function is resolved with the value returned by `onFulfilled`.

The `resolve` method changed here does not currently return a value, this is fine since it's an optional behavior, but not returning a value prevents foreign thenables from keeping the transformed value after going through `Promise.all`

If you have this same thenable object, you can do this:

    p.then((v) => v * 2).then((v) => v * v); // 16

But you can't do this:

    Promise.all([p]).then((result) => store(result[0])); // store is just a function to save the value
    p.then((v) => v * 2); // NaN because v is undefined

benjamingr:
This would break compatibility with ES2015 promises.

chuym:
Could you please provide a testcase where it would break?

Just added this and it worked correctly:

    specify("should be compatible with ES2015 promises", function() {
        var bluebird = Promise.noConflict();
        var p = new Promise(function (resolver) {
            resolver(2);
        });
        global.Promise = bluebird;

        return Promise.all([p, p]).then(
            function (results) {
                assert.equal(results[0], results[1]);
            });
    });

In fact, without the change, the test fail for ES5 promises.

Also:

> Promise.prototype.then(onFulfilled, onRejected)
Appends fulfillment and rejection handlers to the promise, and returns a new promise resolving to the return value of the called handler, or to its original settled value if the promise was not handled (i.e. if the relevant handler onFulfilled or onRejected is undefined).

The handler in bluebird is not returning anything (which is fine, but returning something is also fine as per the standard). Not returning anything prevents from chaining `then` methods because the returned promise from the previos `then` is resolved to `undefined`.

petkaantonov:
[ES spec](http://www.ecma-international.org/ecma-262/6.0/#sec-promise-resolve-functions) explicitly says `return undefined` in the last step.

chuym:
@petkaantonov You linked to the `Promse.resolve`, not the `onFulfilled` callback, it is not the same thing changed here. I believe we are all getting confused here, my apologies if I'm not explaining myself properly here, this is a real issue, please read on and let me know if I'm still incorrect.

The change done on this `PR` is done to the `resolve` function [here](https://github.com/petkaantonov/bluebird/blob/master/src/thenables.js#L70), and as you can see, this function is passed to the `then` function of the thenable object [here](https://github.com/petkaantonov/bluebird/blob/master/src/thenables.js#L62), even though the function is named `resolve`, is is *not* the resolve function you linked (this may be the source of the confusion here), it is analogous to the `onFulfilled` function defined on the [ES spec](http://www.ecma-international.org/ecma-262/6.0/#sec-performpromisethen) and it *mandates* to return something, if it is not d

In the case of `Promise.all`, an explicit `onFulfilled` function is not provided for each promises passed in the iterable, this means that the `onFulfilled` must be the `Identty` function, that is, it must return the value passed to it which is not doing:

    function resolve(value) {
        if (!promise) return;
        promise._resolveCallback(value);
        promise = null;
    }

Even though my change returns `value` only if `promise` is truthy, I believe the correct implementation should always return `value` in order to comply with the ES specifiaction, something like:

    function resolve(value) {
        if (!promise){
            promise._resolveCallback(value);
            promise = null;
        }
        return value;
    }

benjamingr:
I'm confused here @chuym why would it matter if `onFulfilled` returns `value` or not? 

Can you please show in a short example the change in behavior this introduces?

petkaantonov:
'onFulfilled' is user function passed to the promise implementation, the function passed to a thenable as first argument from the implementation is resolve and must return undefined.

chuym:
@benjamingr Returning a `value` from `onFulfilled` allows for the returned `Promise` from the `then` method be resolved with that `value`.
Here:
http://www.ecma-international.org/ecma-262/6.0/#sec-promisereaction-records
> The function that should be applied to the incoming value, and whose return value will govern what happens to the derived promise. If [[Handler]] is "Identity" it is equivalent to a function that simply returns its first argument. If [[Handler]] is "Thrower" it is equivalent to a function that throws its first argument as an exception.

Since the current handler used in `Promise.all` does not return anything, then the returned promise from `then` is given the value `undefined`.
````javascript
var promise = new Promise(function (resolve) {
    resolve(5);
}); // Where Promise is any class that implements the ES5 standard

promise.then(function (value) {
    console.log(value); // 5
    return value * 2;
}).then(function (value) {
    console.log(value); // 10;
}).then(function (value) {
    console.log(value); // undefined (we didn't return anything on the previous then.
});
````
The above behaves as you may expect, and it behaves like that for both Bluebird Promises and ES5 promises or other alternate Promise implementation that follows the standard.

Now when passing these promises to `Bluebird.all`

````javascript
var first = new Promise(function (resolve) {
    resolve(5);
}); // Where Promise is any class that implements the ES5 standard

var second = new Promise(function (resolve) {
    resolve(20);
}); // Where Promise is any class that implements the ES5 standard

var third = new Bluebird(function (resolve) {
    resolve(50);
});

Bluebird.all([first, second, third]).then(function (results) {
    console.log(results); // [5, 20, 50];
});

first.then(function (value) {
    console.log(value); // undefined
});

second.then(function (value) {
    console.log(value); // undefined
});

third.then(function (value) {
    console.log(value); // 50 (this works because for bluebird promises, an alternate behavior is done
});
````

`Bluebird.all` calls the `then` method for every promise, that is not a `Bluebird` promise, with a `onFulfilled` handler that doesn't return anything, thus making the returned promise from `then` to be resolved to `undefined` preventing further `then` methods to be called on the promises passed to `Bluebird.all`

chuym:
@petkaantonov It's the other way around. `resolve` is a function passed to the `Promise` constructor which must return undefined as you say, while `onFulfilled` is a function passed to `then` method that governs what the settled value is for the returned `Promise` from the `then` call.

petkaantonov:
The promise implementation passes the `resolve`, `reject` functions to a thenable's then function as arguments. The thenable's then method then calls them or not, but when it calls them, they must return undefined which the spec is explicit about. You are asking for them to return non undefined.

petkaantonov:
1. What is the ES5 standard? Promises are introduced first in ES6/ES2015.

2. Do you have jsfiddle for your example? `Promise` implementation in this example is horribly broken if it's resolved with a value like 20 and thenning it gives you undefined.

benjamingr:
@chuym that's not how promises behave at all. 

https://jsfiddle.net/4z8q337f/1/

Promises are immutable, once their value changes it should remain the same. If they return 20 once, they will return 20 forever.

PR #936, opened 2015-12-28
Fixed incorrect order of arguments

opensourcejunkie:


opensourcejunkie:
Can

opensourcejunkie:
Sorry, accidentally closed.

benjamingr:
The order is correct, it changed between 2.x and 3.x

PR #937, opened 2015-12-29
Fix a typo

mgenware:
The example is written in Promise.mapSeries, not Promise.each


benjamingr:
Not a typo, but I guess it is a better analogy - thanks!

PR #940, opened 2015-12-30
Fix links to explicit construction anti-pattern.

jmm:
Supercedes #901. As mentioned there I still think it likely makes sense to consolidate by linking to the website from https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns instead of duplicating the content there.


benjamingr:
Thanks!

jmm:
Thanks!

PR #941, opened 2016-01-02
New function: collect

carli2:
Sometimes, one want simply fire some requests and continue only with those values that resolved, but only after all promises either were rejected or resolved.

.collect works as follows:

```Promise.collect([promises, ....]).then(function (result) {
  // result now contains all values of those promises that resolved
});```

Please leave a comment if this feature will be merged. If so, please tell me what is left to do besides API and docs (ammount of tests etc.)


benjamingr:
No, this belongs in a separate library

benjamingr:
Is this acceptable (separate library) @carli2

carli2:
I think, when I need this functionality, I will always implement it on my own.

benjamingr:
@carli2 I was suggesting releasing a library. That's fine though.

PR #944, opened 2016-01-06
Backport disposer docs and a few other enhancements

spion:
Not sure if I should also make all Promise.using references links...

### DIFF
--- a/docs/docs/api/disposer.md
+++ b/docs/docs/api/disposer.md
@@ -15,9 +15,9 @@ title: .disposer
 
 A meta method used to specify the disposer method that cleans up a resource when using `Promise.using`.
 
-Returns a Disposer object which encapsulates both the resource as well as the method to clean it up. 
-The user can pass this object to `Promise.using` to get access to the resource when it becomes available,
-as well as to ensure its automatically cleaned up.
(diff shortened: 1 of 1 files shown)


benjamingr:
Awesome. I'd like to see some more common use cases like in the promisify examples.

docs/docs/api/disposer.md
```diff
@@ -13,8 +13,11 @@ title: .disposer
 .disposer(function(any resource, Promise usingOutcomePromise) disposer) -> Disposer
 ```
 
+A meta method used to specify the disposer method that cleans up a resource when using `Promise.using`.
 
-A meta method used to specify the disposer method that cleans up a resource when using undefined.
+Returns a Disposer object which encapsulates the scarce resource as well as the method to clean it up.
```
tefla: Not sure 'scarce' is required here.

tefla:
:+1: Much needed update to a great feature, but was very difficult to use without looking at the source code. Thanks @spion

docs/docs/api/disposer.md
spion: True, by itself it doesn't say enough. I'd like to expand on the kind of resources which are typically managed this way, but for now I think just removing it will suffice.

spion:
@benjamingr i don't know of any typical use cases other than SQL connections (mine are kinda domain-specific). IMO promisify benefits from more examples as there is a lot of variation for it to deal with - disposers can only really be used in one way

PR #961, opened 2016-01-14
Add longStackTraces to API Reference

chetbox:
I found `Promise.longStackTraces` using Google but noticed it wasn't on the API Reference page. I've added it under "Utility" given that's the section that seems to make the most sense.


benjamingr:
longStackTraces is deprecated. Use Promise.config instead

benjamingr:
Thanks for the contribution anyway by the way. Docs help is really appreciated.

PR #962, opened 2016-01-15
Docs: Added deprecation notice to Promise.longStackTraces

chetbox:
In response to #961 I've added a deprecation notice to the `Promise.longStackTraces` page for anyone who lands there in future.


chetbox:
Is it worth listing deprecated APIs on the `api-reference` page for clarity? I'd suggest something like this:

~~[Promise.longStackTraces](http://bluebirdjs.com/docs/api/promise.longstacktraces.html)~~

benjamingr:
Awesome, thanks!
I think it's better not to include it in the API reference .

PR #991, opened 2016-02-03
Minor updates to resource-management docs

justinhelmer:
- change `getConnection()` to `connectionPool.getConnectionAsync()` for better symmetry
- prefer array format of `using()` for tighter juxtaposition with `all()`
- match formatting of example it is being compared with

### DIFF
--- a/docs/docs/api/resource-management.md
+++ b/docs/docs/api/resource-management.md
@@ -35,14 +35,18 @@ this would lose a lot of the benefits of using asynchronity and we might almost 
 We can do better, retaining concurrency and not leaking resources, by using `.using`:
 
 ```js
-var using = Promise.using;
-
-using(getConnection(),
-      fs.readFileAsync("file.sql", "utf8"), function(connection, fileContents) {
-    return connection.query(fileContents);
-}).then(function() {
-    console.log("query successful and connection closed");
-});
+function doStuff() {
+    return Promise.using([
+        connectionPool.getConnectionAsync(),
+        fs.readFileAsync("file.sql", "utf8")
+    ]), function(connection, fileContents) {
+        return connection.query(fileContents);
+    }).then(function() {
+        console.log("query successful and connection closed");
+    }).finally(function() {
+        connection.close();
+    });
+}
 ```
 </markdown></div>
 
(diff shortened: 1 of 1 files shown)


justinhelmer:
Got a little tripped up on the example, and if my understanding is accurate and these adjustments correct, I believe these minor changes may help others in the future.

justinhelmer:
Use [?w=1](https://github.com/petkaantonov/bluebird/commit/8748f5974530640b794e66d15463397820da55c1?w=1) to see the _real_ diff.

docs/docs/api/resource-management.md
```diff
@@ -35,14 +35,18 @@ this would lose a lot of the benefits of using asynchronity and we might almost
...
-      fs.readFileAsync("file.sql", "utf8"), function(connection, fileContents) {
-    return connection.query(fileContents);
-}).then(function() {
-    console.log("query successful and connection closed");
-});
+function doStuff() {
+    return Promise.using([
+        connectionPool.getConnectionAsync(),
+        fs.readFileAsync("file.sql", "utf8")
+    ]), function(connection, fileContents) {
```
spion: The array format doesn't quite work that way. It will pass a single argument to the function - an array of allocated resources. 

Its the intention that the array variant should be used with an array of (same or similar) resources. Bluebird's `Promise.using` is actually more similar to `Promise.join`. I'd recommend using the non-array variant for a fixed number of heterogenous resources (one of the reasons being that its much easier to write type definitions for those).

docs/docs/api/resource-management.md
```diff
@@ -35,14 +35,18 @@ this would lose a lot of the benefits of using asynchronity and we might almost
...
+        return connection.query(fileContents);
+    }).then(function() {
+        console.log("query successful and connection closed");
+    }).finally(function() {
+        connection.close();
```
spion: Its not necessary to dispose of resources with `Promise.using`. As soon as the promise returned within the closure passed to `using` resolves, the disposer for the connection will be immediately invoked. `Using` will then wait for it to call the `close()` method (waiting for any promises it might return), then return the original result.

Basically `Promise.using` is a safe, automated way to ensure that finalizers are called. 

The docs may not be quite clear in stating this - hopefully we can improve them together.

justinhelmer:
Thanks for following up; I learned a ton in the last five minutes.

Cheers

PR #992, opened 2016-02-04
make tests work when run in a child_process

evanlucas:
`process.stdout.cursorTo` and `process.stdout.clearScreenDown` only exist 
if stdout is a tty. When smoke testing Node.js builds, [citgm](https://github.com/nodejs/citgm) is used 
to make sure breaking changes are not being made to popular packages.
This package fails due to not checking if the `cursorTo` and
`clearScreenDown` functions exists on process.stdout before using them.

(https://github.com/nodejs/citgm/issues/70)

/cc @TheAlphaNerd

### DIFF
--- a/tools/test.js
+++ b/tools/test.js
@@ -246,8 +246,8 @@ var testResults = Promise.join(tests, buildResult, function(tests) {
     } else if (singleTest) {
         return runTestGroup(tests, options);
     } else {
-        process.stdout.cursorTo(0, 0);
-        process.stdout.clearScreenDown();
+        utils.cursorTo(0, 0);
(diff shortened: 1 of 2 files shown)


petkaantonov:
There was checks for them in 3.2.0 but I had to quickly revert it due to a critical issue and haven't had time to add them back yet.

TheAlphaNerd:
@petkaantonov as soon as those checks are back in we can add bluebird to citgm, the smoketester we use before every release

petkaantonov:
Ok landed this just now in 3.2.2.

evanlucas:
Thanks @petkaantonov!

PR #994, opened 2016-02-05
doc: Fix the "global rejection events" links

kevinoid:
It appears that the previously-relative "global rejection events" links on the `Promise.onPossiblyUnhandledRejection` and `Promise.onUnhandledRejectionHandled` pages were overlooked when the API documentation was split into separate pages in 94fd4fe6.  This PR fixes the links to point to the correct page.

Best,
Kevin


spion:
Thank you!

kevinoid:
My pleasure.  Thanks for the quick merge!

PR #1001, opened 2016-02-09
Make var promisification explicit in the docs

baer:
The docs imply that `Promise.promisify(<anything>)` will manipulate instances of that object that are already imported.


petkaantonov:
It adds methods to the class so all instances are indeed affected

PR #1002, opened 2016-02-10
FIXES #1000

### LINKED ISSUE #1000: Promise.delay will not clearTimeout when cancelled
Using `bluebird 3.2.2` 
Similar to  #926 with `timeout` When canceling a `Promise.delay` the internally used timeout is not cleared and the process will hang

```javascript
Promise.config({cancellation: true});

process.on('exit', function() {
  console.log(new Date(), 'Process exiting');
});

console.log(new Date(), 'Starting');
var delayPromise = Promise.delay(10000);

delayPromise.finally(function () {
	if (delayPromise.isCancelled()) {
		console.log(new Date(), 'Cancelled');
	}
});

setTimeout(function() {
	console.log(new Date(), 'Cancelling promise');
	delayPromise.cancel();
}, 1000);
```
Output -
```
Tue Feb 09 2016 05:17:03 GMT+0000 (UTC) 'Starting'
Tue Feb 09 2016 05:17:04 GMT+0000 (UTC) 'Canceling promise'
Tue Feb 09 2016 05:17:04 GMT+0000 (UTC) 'Cancelled'
Tue Feb 09 2016 05:17:13 GMT+0000 (UTC) 'Process exiting'
```

Notice the 10 seconds delay between the promise being canceled and the process exiting

yonjah:
Add clearTimeout on delay cancel

### DIFF
--- a/src/timers.js
+++ b/src/timers.js
@@ -31,6 +31,9 @@ var delay = Promise.delay = function (ms, value) {
     if (value !== undefined) {
         ret = Promise.resolve(value)
                 ._then(afterValue, null, null, ms, undefined);
+        if (debug.cancellation() && value instanceof Promise) {
+            ret._setOnCancel(value);
+        }
     } else {
         ret = new Promise(INTERNAL);
         handle = setTimeout(function() { ret._fulfill(); }, +ms);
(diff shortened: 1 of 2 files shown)


petkaantonov:
Thank you. The case where `value` is should be handled as well.

yonjah:
I think it's OK now

petkaantonov:
I don't understand how in the first case the timeout is cleared after cancel, it doesn't even refer to `handle`

yonjah:
Do you mean, when delay is called with a value ?
The timeout isn't cleared in that case since it's not being set yet.
It will be set on the next call to delay (being made from `afterValue`) and in that case a onCancel handler is set as well.

petkaantonov:
Landed in 3.3

PR #1010, opened 2016-02-17
Update timers.js

eldechu:
Adding call stack for message when timeout occurs


petkaantonov:
You can enable long stack traces and use .stack of the TimeoutError.

eldechu:
Isn't that so "Long stack traces imply a substantial performance penalty, around 4-5x for throughput and 0.5x for latency." ? http://bluebirdjs.com/docs/api/promise.longstacktraces.html
I'm interrested to have error stack trace for timeout not only for debug, but also for runtime.

petkaantonov:
This would incur same performance penalty for all users regardless of their configuration

PR #1016, opened 2016-02-24
Fix INLINE_SLICE when arguments.length < 1

wltsmrz:
The current generated code for INLINE_SLICE in the case of `promisify()` looks like

```js
var $_len = arguments.length;
var args = new Array($_len - 1);
for (var $_i = 1; $_i < $_len; ++$_i) {
  args[$_i - 1] = arguments[$_i];
};
```
multiArgs check for arguments.length > 2 was removed from `promisify()` in [this commit](https://github.com/petkaantonov/bluebird/commit/6f4863015db149570990ba3a184fcc02691aaf48). Now, this will throw a RangeError if no arguments are given to the callback in `promisify()`. (`new Array(-1)`). Proposed INLINE_SLICE looks like

```js
var $_len = arguments.length;
var args = [];
for (var $_i = 1; $_i < $_len; ++$_i) {
  args[$_i - 1] = arguments[$_i];
};
```

Benchmarks seem more or less unaffected. Of course, at least one (error) argument *should* be provided to the callback, but I see no harm in defaulting to empty array. If it's decided to enforce the at-least-one-argument rule, I'll update the pull request for throwing a human-friendly error instead

### DIFF
--- a/docs/docs/api/catch.md
+++ b/docs/docs/api/catch.md
@@ -40,10 +40,10 @@ Example:
 somePromise.then(function() {
     return a.b.c.d();
 }).catch(TypeError, function(e) {
-    //If a is defined, will end up here because
+    //If it is a TypeError, will end up here because
     //it is a type error to reference property of undefined
 }).catch(ReferenceError, function(e) {
-    //Will end up here if a wasn't defined at all
+    //Will end up here if a was never declared at all
 }).catch(function(e) {
     //Generic catch-the rest, error wasn't TypeError nor
     //ReferenceError


petkaantonov:
Shouldn't `Math.max(0, $len - 1)` be enough? `[]` is same as `new Array(4)` and setting .length to 0, so it wouldn't affect benchmarks with less than 5 arguments.

benjamingr:
Wouldn't not passing an error break promisification anyway?

wltsmrz:
@petkaantonov Is there a significant reason to prefer using array constructor?

@benjamingr The assumption would be that absence of error indicates success, but I would expect defaulting on this assumption to be contentious, so if there's agreement I'll update the pull request to throw a human-friendly error instead. A pattern in particular that I have in mind, which I've seen in the wild:

```js
var net = require('net')

require('bluebird').promisifyAll(net, {multiArgs: true})

net.connectAsync(1234)
.then(() => console.log('connected'))
// RangeError: Invalid array length
```

The callback is an event listener that is given zero arguments.

benjamingr:
> @petkaantonov Is there a significant reason to prefer using array constructor?

It lets the engine pre-allocate the array which is slightly faster.


> @benjamingr The assumption would be that absence of error indicates success, but I would expect defaulting on this assumption to be contentious, so if there's agreement I'll update the pull request to throw a human-friendly error instead. 

That pattern is bad because it implies promisifying an API that does not take regular node callbacks. The problem here is that you have the illusion of knowing when an error happened when in fact it is emitted by an `error` event.

wltsmrz:
@benjamingr [Most of the time](https://github.com/nodejs/node/blob/7b0a83d2b007a381effa038b3e4fb2421de10b87/lib/net.js#L914-L927) that is the case . It seems possible that someone else will run into this pattern even if node/net/connect itself isn't promisifiable. Options:

+ Throw a human-friendly error
+ Assume the best, slice to empty array
+ Ignore the matter, users are left to their own devices with a courtesy RangeError
+ ?

benjamingr:
I'm always +1 on better errors :)

PR #1018, opened 2016-02-24
Fix error on improperly-invoked promisify callback

snowblindroan:
Prior discussion: https://github.com/petkaantonov/bluebird/pull/1016#issuecomment-188247268

### DIFF
--- a/src/nodeback.js
+++ b/src/nodeback.js
@@ -34,6 +34,12 @@ function wrapAsOperationalError(obj) {
 function nodebackForPromise(promise, multiArgs) {
     return function(err, value) {
         if (promise === null) return;
+
+        if (arguments.length < 1) {
+            err = new Error("Promisified function invoked callback with too " +
+                            "few arguments");
+        }
+
         if (err) {
             var wrapped = wrapAsOperationalError(maybeWrapAsError(err));
             promise._attachExtraTrace(wrapped);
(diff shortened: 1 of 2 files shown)


benjamingr:
LGTM

petkaantonov:
But the user cannot do anything about this error... I think if there is no arguments then it should resolved with undefined? Or what is the intention of the callback when it calls with 0 arguments?

benjamingr:
In 99% of cases it's probably people who promisified event emitters incorrectly.

petkaantonov:
How would promisifying event emitter cause a callback to be called?

benjamingr:
```
net.connectAsync().then(() => {
    
});
```

petkaantonov:
`connectListener` is that really called with 0 arguments? isn't it called with socket?

benjamingr:
That's a good point, I recall it was in the 0.8 happy days when I was still writing to raw tcp.

wltsmrz:
I don't recall that it ever was. As far as I can tell the callback to `createConnection()` was introduced in [this commit](https://github.com/nodejs/node/commit/c60cdcda4e39081fa56568df941aff1a4b56b39c). There is [no argument](https://github.com/nodejs/node/blob/c60cdcda4e39081fa56568df941aff1a4b56b39c/lib/net_uv.js#L351-L359) to the callback

benjamingr:
Oh yeah, I'm thinking about createServer my bad.

benjamingr:
@petkaantonov why the close?

petkaantonov:
Errback called with no arguments can also be a success, can't it?

benjamingr:
Oh yeah, I've never seen this though.

PR #1035, opened 2016-03-05
Fix broken link in docs

RebootJeff:


RebootJeff:
I'm guessing the static site builder has trouble creating names for header anchors that have parens?

benjamingr:
thanks

PR #1044, opened 2016-03-16
Update async benchmark

suguru03:
[Async.js](https://github.com/caolan/async#forEachOf) was added `forEachOf` after v1.x.
The function supports parallel execution and getting array index.
Furthermore, the function is faster than `async.parallel`.
Could you check this PR?

### DIFF
--- a/benchmark/madeup-parallel/callbacks-caolan-async-parallel.js
+++ b/benchmark/madeup-parallel/callbacks-caolan-async-parallel.js
@@ -1,22 +1,13 @@
 require('../lib/fakes');
 var async = require('async');
 
-function fileInsertFor(i, tx) {
-    return function(callback) {
-        FileVersion.insert({index: i})
-            .execWithin(tx, callback);
-    };
-}
-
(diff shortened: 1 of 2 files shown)


petkaantonov:
>Like each, except that it iterates over objects, and passes the key as the second argument to the iteratee.

suguru03:
How about [`async.times`](https://github.com/caolan/async#times)? https://github.com/suguru03/bluebird/commit/358b542b27ba45fc43350ca9105e2686bea4ec21
It is similar to [`madeup-parallel`](https://github.com/petkaantonov/bluebird/blob/master/benchmark/madeup-parallel/callbacks-baseline.js).
It supports parallel execution too and is faster than `async.parallel`.
Could you tell me your opinion?

petkaantonov:
index is just an accident of the artificial benchmark, in reality you wouldn't have the queries be based on index and you wouldn't be able to use .times like this.

suguru03:
Thank you very much for your opinion.

PR #1046, opened 2016-03-17
Fix typo in PromiseInspection API documentation.

MattiasBuelens:
`PromiseInspection.pending()` does not exist, should be `PromiseInspection.isPending()`


benjamingr:
Thanks!

PR #1061, opened 2016-03-30
Copy Esailija's answer from stackexchange

jghaines:
Cut, paste, edit from:
Why are native ES6 promises slower and more memory-intensive than bluebird?
http://programmers.stackexchange.com/a/279003/24609


benjamingr:
I say we merge - @petkaantonov ?

petkaantonov:
The article answers the question why you should care about promise performance, not why bluebird is fast

PR #1064, opened 2016-04-06
Prettify doccument

nhducit:
Add an empty line before code tag ( ```js ).

http://bluebirdjs.com/docs/error-explanations.html#error-expecting-an-array


benjamingr:
It looks just fine the way it is, there are no spacing issues..

PR #1067, opened 2016-04-11
Update catch.md

vladimirkosmala:
Added the fact that one can return a value or a promise from a catch() to recover from a failure.


benjamingr:
This gets raised up a lot, so sure. I'll merge and edit.

PR #1075, opened 2016-04-18
Fix scheduler not being called when set during queue draining

janpaul123:
We had a failing test in our test suite at @remix, which I was able to
track down to the scheduler not being called from async.js, because
_isTickUsed was not being set to false until after draining the queue.

I don't have time now to write a test case for this, so hopefully
someone can figure that out.

It might be helpful to note that we're using a custom scheduler that
uses `setTimeout(fn, 0)`, but where `setTimeout` is mocked by Jasmine so
that we have control over asynchronous operations.

### DIFF
--- a/src/async.js
+++ b/src/async.js
@@ -153,8 +153,8 @@ Async.prototype._drainQueue = function(queue) {
 
 Async.prototype._drainQueues = function () {
     ASSERT(this._isTickUsed);
-    this._drainQueue(this._normalQueue);
     this._reset();
+    this._drainQueue(this._normalQueue);
     this._haveDrainedQueues = true;
     this._drainQueue(this._lateQueue);
 };


benjamingr:
@janpaul123 hey, mind elaborating on why you're using `setTimeout` as a scheduler?

janpaul123:
Yes, because it gets mocked by jasmine.clock so in our tests we don't have asynchronous calls, which can lead to pollution of state between tests if a call is executed after a test has finished.

petkaantonov:
The scheduler must be asynchronous. If you are using mocha you can return promises from test functions that are awaited for before the next test begins.

benjamingr:
@petkaantonov any chance we can warn against this?

petkaantonov:
@benjamingr we *are* warning against it... see the *danger*: http://bluebirdjs.com/docs/api/promise.setscheduler.html

benjamingr:
I meant via an actual warning

petkaantonov:
Doesn't seem there is need for a runtime warning, it would affect performance too. The documentation is quite clear about it, it's an advanced function only needed once per application and the misues appears to be rare.

janpaul123:
@petkaantonov: I think you're misunderstanding what I said on several points:

1. The scheduler is asynchronous, just in the same tick. I'm not setting the scheduler to `(fn) => fn()` or something, no, it's just mocked by Jasmine, but it will still behave similar to a true asynchronous scheduler. The scheduler will be called after all other code has been executed.

2. I know how to wait for promises in functions under test, obviously. The problem that I'm trying to solve for is if the code under test has some side effect that happens in a promise, and that the test writer didn't know about (maybe it was added later, etc). In that case you don't want that code to execute while the test has already finished.

But I see now how this PR isn't the solution either, and I was mistaken about what the problem was, and I still don't quite know.

PR #1072, opened 2016-04-16
Add documentation of `Promise.coroutine()` `yieldHandler` option

overlookmotel:
This PR adds documentation for `Promise.coroutine()` `yieldHandler` option.

It also simplifies one of the other code examples for `Promise.coroutine.addYieldHandler()`.

PR #1076, opened 2016-04-18
Static method to create separate instance of Bluebird

overlookmotel:
This PR is intended as a basis for discussion, rather than "merge me now!"

It feels to me that in many cases module authors should be encouraged to use their own instance of bluebird, rather than the "global" bluebird retrieved by `require('bluebird')`. Here are some examples of where this is the case:

1. Module modifies/shims Bluebird in some way e.g. [cls-bluebird](https://www.npmjs.com/package/cls-bluebird), [bluebird-extra](https://www.npmjs.com/package/bluebird-extra). The host app or other modules may not expect these modifications and it may break them. In the case of [cls-bluebird](https://www.npmjs.com/package/cls-bluebird), it reduces performance of bluebird globally.

2. Bluebird constructor is altered by using static methods/properties of Bluebird constructor. e.g. I am writing a module in which I wish to enable cancellation, but it would be a bad thing if this also enables cancellation in the app that requires my module, as the app author probably wouldn't expect this and it could cause unexpected behaviour.

It is already possible to create a new independent instance of Bluebird with:

```js
require('bluebird/js/release/promise')() // bluebird v3.x
require('bluebird/js/main/promise')() // bluebird v2.x
```

However, this has a few problems:

1. The code is different for v2.x and v3.x
2. It's not accessible if bluebird is injected into a module rather than required directly
3. It may not be possible in the browser (I don't know - I'm server-side only)
4. It's a bit obtuse - hard to encourage people to use it.

This PR adds a static method `.newBluebird()` on the Bluebird constructor, which returns a new independent instance of Bluebird.

```js
var Promise2 = Promise.newBluebird();
assert.notEqual(Promise, Promise2);
```

There may be better ways to achieve the same thing, and better names for the method than `newBluebird`, but I think in principle it's a good idea for Bluebird to provide this functionality with an easy-to-access method.

What do you think?

### DIFF
--- a/docs/docs/api-reference.md
+++ b/docs/docs/api-reference.md
@@ -74,6 +74,7 @@ redirect_from: "/docs/api/index.html"
     - [.catchReturn](api/catchreturn.html)
     - [.catchThrow](api/catchthrow.html)
     - [.reflect](api/reflect.html)
+    - [Promise.getNewLibraryCopy](api/promise.getnewlibrarycopy.html)
     - [Promise.noConflict](api/promise.noconflict.html)
     - [Promise.setScheduler](api/promise.setscheduler.html)
 - [Built-in error types](api/built-in-error-types.html)
--- a/docs/docs/api/promise.getnewlibrarycopy.md
+++ b/docs/docs/api/promise.getnewlibrarycopy.md
@@ -1,0 +1,64 @@
+---
+layout: api
+id: promise.getnewlibrarycopy
+title: Promise.getNewLibraryCopy
+---
+
+
+[← Back To API Reference](/docs/api-reference.html)
+<div class="api-code-section"><markdown>
+##Promise.getNewLibraryCopy
+
+```js
+Promise.getNewLibraryCopy() -> Object
+```
+
+Returns a new independent copy of the Bluebird library.
+
+This method should be used before you use any of the methods which would otherwise alter the global `Bluebird` object - to avoid polluting global state.
+
+A basic example:
+
+```js
+var Promise = require('bluebird');
+var Promise2 = Promise.getNewLibraryCopy();
+
+Promise2.x = 123;
+
+console.log(Promise2 == Promise); // false
+console.log(Promise2.x); // 123
+console.log(Promise.x); // undefined
+```
+
+`Promise` is independent to `Promise`. Any changes to `Promise2` do not affect the copy of Bluebird returned by `require('bluebird')`.
+
+In practice:
+
+```js
+var Promise = require('bluebird').getNewLibraryCopy();
+
+Promise.coroutine.addYieldHandler( function() { /* */ } ); // alters behavior of `Promise.coroutine()`
+
+// somewhere in another file or module in same app
+var Promise = require('bluebird');
+
+Promise.coroutine(function*() {
+    // this code is unaffected by the yieldHandler defined above
+    // because it was defined on an independent copy of Bluebird
+});
+```
+</markdown></div>
+
+<div id="disqus_thread"></div>
+<script type="text/javascript">
+    var disqus_title = "Promise.getNewLibraryCopy";
+    var disqus_shortname = "bluebirdjs";
+    var disqus_identifier = "disqus-id-promise.getnewlibrarycopy";
+
+    (function() {
+        var dsq = document.createElement("script"); dsq.type = "text/javascript"; dsq.async = true;
+        dsq.src = "//" + disqus_shortname + ".disqus.com/embed.js";
+        (document.getElementsByTagName("head")[0] || document.getElementsByTagName("body")[0]).appendChild(dsq);
+    })();
+</script>
+<noscript>Please enable JavaScript to view the <a href="https://disqus.com/?ref_noscript" rel="nofollow">comments powered by Disqus.</a></noscript>
--- a/src/promise.js
+++ b/src/promise.js
@@ -165,6 +165,8 @@ Promise.prototype.error = function (fn) {
     return this.caught(util.originatesFromRejection, fn);
 };
 
+Promise.getNewLibraryCopy = module.exports;
+
 Promise.is = function (val) {
     return val instanceof Promise;
 };
--- a/test/mocha/getNewLibraryCopy.js
+++ b/test/mocha/getNewLibraryCopy.js
@@ -1,0 +1,30 @@
+"use strict";
+
+var assert = require("assert");
+var testUtils = require("./helpers/util.js");
+
+describe("Promise.getNewLibraryCopy", function() {
+    it("should return an independent copy of Bluebird library", function() {
+        var Promise2 = Promise.getNewLibraryCopy();
+        Promise2.x = 123;
+
+        assert.equal(typeof Promise2.prototype.then, "function");
+        assert.notEqual(Promise2, Promise);
+
+        assert.equal(Promise2.x, 123);
+        assert.notEqual(Promise.x, 123);
+    });
+    it("should return copy of Bluebird library with its own getNewLibraryCopy method", function() {
+        var Promise2 = Promise.getNewLibraryCopy();
+        var Promise3 = Promise2.getNewLibraryCopy();
+        Promise3.x = 123;
+
+        assert.equal(typeof Promise3.prototype.then, "function");
+        assert.notEqual(Promise3, Promise);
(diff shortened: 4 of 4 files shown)


benjamingr:
> Bluebird constructor is altered by using static methods/properties of Bluebird constructor. e.g. I am writing a module in which I wish to enable cancellation, but it would be a bad thing if this also enables cancellation in the app that requires my module, as the app author probably wouldn't expect this and it could cause unexpected behaviour.


Note that 3.x cancellation isn't prone to this problem given its disinterest semantics. This was an issue in 2.x though.

I generally agree that module authors are usually better off getting their own bluebird and configuring it/extending it to be useful in their module if they're doing anything but using promises for wrapping libraries. 

I just don't think that this simplifies it over the current `require` way in particular. (It's possible in the browser by the way). 

I'm fine with `require("bluebird/new")` or something like that though.

overlookmotel:
@benjamingr Thanks for swift reply.

Yes, I agree that v3.x cancellation is much less problematic in this regard. However, in some weird cases cancellation being enabled in one module *might* affect behaviour in another. If there's `promise.cancel()` in one module (maybe some errant code that shouldn't be there), then tests all pass fine with cancellation turned off as `cancel()` is a no-op in that case - but then another module switches cancellation on, and it breaks.

But that's me nit-picking.

The case I mention above where Bluebird Promise constructor is injected into a module, rather than required within the module still stands - you can't then use `require('bluebird/new')` within the module. Though there are probably work-arounds for that.

But I suppose my more general point is that if module authors using their own instance of Bluebird is to be encouraged, it's good to make that facility visible and as easy as possible. Adding a method to the API seems to me the simplest way to do that.

Does adding a method to the constructor create performance overhead?

PS I'm working on making Bluebird work better with [cls](https://www.npmjs.com/package/continuation-local-storage) at the moment ([cls-bluebird](https://www.npmjs.com/package/cls-bluebird) is broken for bluebird v3.x) and I'll probably have some more issues to raise here soon. I hope I don't become a pain.

overlookmotel:
@benjamingr Have you had any more thoughts on this?

overlookmotel:
Ping @benjamingr . Sorry to hassle you, but have you had time to think about what approach you'd prefer on this?

benjamingr:
Ping @petkaantonov

spion:
You would have a +1 from me if I knew a real case where this was necessary (all the gory details are fine). Is it for sequelize plugins?

overlookmotel:
@spion You're *almost* as good 😆.

In truth I don't have a killer use case of something that requires a `.newBluebird()` method and couldn't be accomplished with `require('bluebird/js/release/promise')()`.

The closest I've got was when researching using bluebird with [CLS](https://www.npmjs.com/package/continuation-local-storage). I [wrote something](https://github.com/overlookmotel/cls-bluebird-test) that tested various different shims that add CLS support to bluebird against - and ran them against different versions of bluebird.

To load 2 different versions of bluebird simultaneously, I had to create two modules [bluebird2](https://www.npmjs.com/package/bluebird2) and [bluebird3](https://www.npmjs.com/package/bluebird3), each of which just exports v2 or v3 of bluebird.

Now I wanted to apply a range of different CLS-bluebird shims on these instances of bluebird, and check the results of each. Obviously I needed a fresh instance of bluebird each time so that I wasn't putting one shim on top of another.

Because `bluebird` was required within another module, rather than directly, `require('bluebird/js/release/promise')` wasn't available.

My workaround was [bluebird2](https://www.npmjs.com/package/bluebird2) and [bluebird3](https://www.npmjs.com/package/bluebird3) add a `.clone()` method onto the instances of bluebird they return, those methods in turn call `require('bluebird/js/release/promise')()`.

However...

My argument for this is less a question of "why?" than "why not?". Here goes:

#### The problem

There's a problem. The problem is that the value of `require('bluebird')` is in a sense a global variable. One module calling `require('bluebird')` can get a reference to the same object as `require('bluebird')` in another module.

And bluebird has some features which alter the behavior of bluebird globally.

The most troublesome of these is `Promise.coroutine.addYieldHandler()`. If one dependency of my app adds a yield handler, that can affect the behavior of `Promise.coroutine()` in other dependencies of my app.

Any app or module author who modifies the global state of bluebird should, in my opinion, create a new instance of bluebird first so they don't pollute the global scope.

But often they don't. Examples are [bluebird-co](https://www.npmjs.com/package/bluebird-co) (adds coroutine yield handler) and [sequelize](https://www.npmjs.com/package/sequelize) (shims bluebird methods to support CLS).

#### Possible solutions

It seems to me that module authors need to be encouraged not to mutate the global `require('bluebird')`.

It's already possible by using `require('bluebird/js/release/promise')()`. But it's rather obscure, and I think that's one reason people don't use it.

Two possible solutions:

1. Add a method `Bluebird.newBluebird()`
2. Create a shortcut `require('bluebird/new')` (as @benjamingr suggested)

#### Why is the `.newBluebird()` approach better?

1. The most effective way to make functionality visible is to add it to the API
2. Helps use cases such as the one above where bluebird is injected rather than directly required
3. Possibly better compatibility with ES6 classes. Can you `import bluebird/new`?

#### And finally... why not?

Does adding a method to the Bluebird constructor create a performance impact? If it doesn't, then what's the harm?

overlookmotel:
PS @spion: your article "Why I am switching to promises" was what introduced me to the wonderful world of promises in the first place. So thanks for that!

spion:
I see what you mean with `addYieldHandler`. That was meant for application authors, not library authors.

Its undocumented, but you can pass the custom yield handler option to Promise.coroutine itself: 

https://github.com/petkaantonov/bluebird/blob/master/src/generators.js#L187-L193

Perhaps we should add it to the docs and clarify that the first one is meant for apps only, second one for libraries...

overlookmotel:
@spion I submitted a PR (#1072) a while back to add `yieldHandler` option to the docs.

NB even if `addYieldHandler` were only used in app code, there's still a possibility that it could have unintended consequences within the application's dependencies that also use `Promise.coroutine`.

But that's by the by... Do you have any thoughts about my general point that it'd be a good thing if module authors used their own instance of bluebird if they're altering bluebird in any way, and that adding a `.newBluebird()` method might help encourage this?

spion:
Yeah I don't see anything wrong with it. We don't have to list it in the main API reference, it could be in the same place as it is now, but as a method... I'm just generally worried about increasing the API surface, if we had a way to deal with that...

petkaantonov:
I don't like the name, it should be something generic related to creating copy of the library, not the "brand" name. But otherwise lgtm.

overlookmotel:
@petkaantonov OK great. Thanks.

How about `.newInstance()` or `.clone()`?

overlookmotel:
@spion I read a bit about monads. Thoroughly boggled my mind. I think I'll leave that for another time...

petkaantonov:
`newInstance` makes me think it creates a new promise instance while `.clone` suggests it clones a promise. How about `.cloneClass`? It seems hard to come up with a name.

benjamingr:
"getNewLibraryCopy"?

You only do it once, no need to make it short.

petkaantonov:
getNewLibraryCopy seems good enough

overlookmotel:
PR updated to rename method to `getNewLibraryCopy`, rebased on latest master and added docs.

overlookmotel:
Tests are failing on node v4 and v5. But is this related to this PR? Or something changing to do with compiling native modules on Travis?

petkaantonov:
the tests somehow pollute the process and need to be ran in separate processes, I added that fix

petkaantonov:
published in 3.4.1

overlookmotel:
Ah OK great. Would you accept a PR for the same to bluebird v2 branch? I for one am stuck on v2 in some apps due to reliance on the old cancellation behaviour.

overlookmotel:
Thanks for merging!

petkaantonov:
yea it could be done to v2 too

jonasjoni:
Wysłane z tabletu Samsung.-------- Oryginalna wiadomość --------Od: Petka Antonov <notifications@github.com> Data: 17.06.2016  22:19  (GMT+01:00) Do: petkaantonov/bluebird <bluebird@noreply.github.com> Temat: Re: [petkaantonov/bluebird] Static method to create separate instance
  of Bluebird (#1076) 
yea it could be done to v2 too

PR #1092, opened 2016-05-06
2 fixes for warning explanation docs

strawbrary:


benjamingr:
How are you finding errors?

Anything else you'd like to see? Are you finding the warnings useful?

Thanks for the PR!

strawbrary:
I updated to Node v6 and I guess something changed that it is now surfacing some errors that were previously swallowed and that's how I found the "Warning: a promise was created in a handler but was not returned from it" one. It was a little confusing that the top of the stack wasn't necessarily the source of the error, but other than that the docs are very helpful so it was relatively painless to figure out what was wrong and fix it.

benjamingr:
Thanks.

PR #1100, opened 2016-05-14
Proposal: `Promise.concurrent(limit)` and `.concurrent(limit)`

justsml:
## Preface
I know concurrency has always been an awkward, if mostly misunderstood feature. (You can't necessarily control how many threads your OS or DB are spinning on. But the importance of being able to **limit the parallel execution of given sub-tasks** is immensely useful ESPECIALLY when you don't want to contend with other system resource. Potentially exhausting resources and leaving everything feeling unresponsive.

## Use Case
I typically configure a `concurrency` limit in my apps to between 2-6 cores - on my 24-core server. I know the sum of **actual spinning threads will exceed** the configured limit, as NodeJS/V8 has it's own background threads, I/O, etc. 
**Without any limits** set, I quickly end up with an **unpredictable & unstable server.**

## Proposed Update

Add `Promise.concurrent(limit)` and `.concurrent(limit)` to make the interface more similar to `Promise.delay()` and `Promise.timeout()` 
The expanded API to provide an alternative to `.map(fn, {concurrency: limit})` is also harder to mis-configure or overlook.

## Example

I've included documentation & updated tests. 

An example adapted from one of my projects looks like this:

```js
var leadData = Promise
    .resolve(fetch('/leads'))
    .concurrency(4)
    .map(lead => {//<- Parallel map (4x)
        return {
            messages: getMessages(lead),
            stats: getStats(lead)
        }
    })
    .concurrency(8)
    .filter(leadData => leadData.messages.length >= 1)//<- Highly parallel (8x)
    .concurrency(1)
    .map(leadData => addToTable(leadData))//<- Not parallel, UI bound (1x)
```

Please let me know if I missed anything or if you have suggestions!

Thanks in advance!!!

- Dan L.


spion:
`map` and `filter` actually have concurrency options:

http://bluebirdjs.com/docs/api/promise.filter.html

justsml:
Passing in wrong options is very possible with the current API. One might also use '...spread' to strange results.
My approach aligns with existing function-based interface. You don't have a 'timeout' option along with 'concurrency' for example

benjamingr:
I think this is confusing as the promises aren't affected by the concurrency limit - only the functions passed are.

I would take a PR that `TypeError`s if a non object is passed.

dhigginbotham:
I like the idea of a chainable fn that allows me to control concurrency limits, would be easier to read than going through another level of config opts per fn that support for concurrency like `map` and `filter`. +1

spion:
timeout takes an existing promise and creates a new promise that resolves when the exiting promise resolves but gets rejected if the specified timeout times out.

I guess a better example would be `bind` but that method has a specific reason to exist (it removes one level of nesting).

Like bind it introduces extra hidden state, which could be error prone i.e. if you forget to unset the concurrency before returning the promise it will affect all subsequent calls w/o the consumer knowing about it.

Overall I'm -1 on it. No significant benefits and there are significant drawbacks.

benjamingr:
To explain further: this would not do what most people would expect based on my experience answering hundreds of questions in Stack Overflow about promises.

One of the most common confusions when people learn promises is understanding that promises are already started operations and you can't "run promises". `map/filter` are borderline as is - even with them, adding `.concurrency` was a really tough call we debated over a year about..

Doing something like `.concurrency(10)` is very misleading. It's the API I "know and love" from C# (in LINQ, when talking about sequences of actions).

justsml:
@spion I understand it introduces side-effects in the local sense. But as you mentioned, so does `.bind`. I respectfully disagree with it having no significant benefits, as I repeatedly have had to review code where Jr. coders have forgot to add a concurrency limit. They often respond to me that it makes no sense to have this "weird option parameter after the method." Also, the number of times i've seen the incorrect `{concurrent: x}` instead of `{concurrency: x}` is kind of amazing.

Let me be clear, I am not suggest removing it as an option.

I'd just like to see a more explicit usage pattern. 

I'm trying to argue for a more homomorphic interface. The 'options argument' design would make more sense if you supported something like the following examples: (which I'm not opposed to, just would like either approach more fleshed out)

```js
Promise.resolve(getData(), {timeout: 2000})
  .map(checkData, {timeout: 1000, concurrency: 4})
  .then(showResults)
// or
getData()
  .filter(expensiveFn, {timeout: 1000, concurrency: 2})
  .then(showResults)

// I feel my design sticks more closely to the functional programming ethos:
getData()
  .timeout(1000)
  .concurrency(2)
  .filter(expensiveFn)
  .then(showResults)
```

As for `.timeout`, where the whole (descending) set of chained calls which fall under the promise instance are time-limited. 
In a similar way, I'd like to restrict all subsequent `.map` or `.filter` calls to have an upper limit on thread use.

------

@benjamingr I really appreciate your examples & responses on StackOverflow - they've certainly helped me immensely over the years. I just was wondering if you could refer/link back to examples of confusion people had regarding concurrency specifically, I've tried to look, but I think I'm only seeing edited answers (missing full picture). 
I'm curious if you've run into anything similar to what I described (above) regarding the challenges I've seen jr. coders run into regarding the syntax?

petkaantonov:
I agree with @benjamingr  this is extremely misleading and confusing, as flawed as option argument is.

spion:
@justsml Its just going to move the problem to "forgot to unset concurrency limit". And you have to read the docs to determine whether it applies only to one subsequent method or to all.

benjamingr:
@justsml 

First of all since we're all being harsh here about your idea I want to start by _thanking you_ for making a contribution and trying to help bluebird. It's not a trivial place to contribute and there are lots of things around to improve.

> I just was wondering if you could refer/link back to examples of confusion people had regarding concurrency specifically, I've tried to look, but I think I'm only seeing edited answers (missing full picture). 


Tons of examples people "running promises":

https://www.google.com/search?q=%22run+promises%22+site%3Astackoverflow.com&oq=%22run+promises%22+site%3Astackoverflow.com&gs_l=serp.3..0i71l2.1199.1199.0.1287.1.1.0.0.0.0.0.0..0.0....0...1c.1.64.serp..1.0.0.D_aztbo8fwo

This is a common source of confusion.

> I'm curious if you've run into anything similar to what I described (above) regarding the challenges I've seen jr. coders run into regarding the syntax?

Yes, I agree - I definitely think we should `throw` if a number is passed as the option rather than an object (and even warn if the object has invalid fields).

I think the concurrency parameter is confusing - it only limits the concurrency of the functions run inside a mapper/filterer function and not in the program. This is why Petka objected to it for over a year. 

I think it's a necessary evil because people have a hard time implementing it correctly in userland and a lot of very bad userland implementations exist.
 
I think tying it to a single function is generally a good idea.

justsml:
@petkaantonov I know you probably cringe in horror every time someone brings up concurrency, sorry.
But I was wondering if you were more horrified at my PR or the examples in my comment https://github.com/petkaantonov/bluebird/pull/1100#issuecomment-219542276 ... I'm just curious, as scoped timeouts for intense `.map` operations might be desirable, but idk.

@benjamingr Thanks for the less-harsh reply. 😉 I absolutely love bluebird - as you can see from this link https://www.google.com/search?q=bluebird+justsml I've been a shameless promoter for ages. I'm sure the core NodeJS ppl are tired of hearing me mention it 😺 

I'll put together a PR to strictly check the `{concurrency}` option. It should be as simple as a check on the options object's keys, roughly:
```js
typeof opt === 'object' && (Object.keys(opt).length <= 1 && Object.keys(opt).indexOf('concurrency') === 0)
```

Let me know if i'm missing something.

Thx.

justsml:
I see @petkaantonov beat me to the options error, thx man.

PR #1126, opened 2016-06-09
Fix logic error in .value|.reason docs

jmm:
```diff
- where it's guaranteed
+ where it's not guaranteed
```


petkaantonov:
Hmm it should say settled in both, it is a programming error to use these methods if the promise isn't fulfilled or rejected

jmm:
Ok, I see, thanks. I was wondering, did you mean to close this or just comment on it?

PR #1131, opened 2016-06-14
Update !== to (apparently) intended ===

cgamache:
Just an oversight I'm sure.


benjamingr:
yup, thanks

benjamingr:
@petkaantonov doc rebuild needed

PR #1134, opened 2016-06-15
fixed wrong test description

Yevs:
There was test description and test method mismatch in `tests/mocha/map.js`


benjamingr:
Thanks

Issue #1136, opened 2016-06-18
`Promise.version` is incorrect for v3.4.1


overlookmotel:
Please answer the questions the best you can:

1) What version of bluebird is the issue happening on?

v3.4.1

2) What platform and version? (For example Node.js 0.12 or Google Chrome 32)

Node 6.2.1, OS X 10.9.5

3) Did this issue happen with earlier version of bluebird?

No.

The value of `Promise.version` in v3.4.1 is `'3.4.0'`. Not a big deal, just I happened to notice it.

petkaantonov:
Fixed since 3.4.2

PR #1156, opened 2016-07-08
Update Promise.version as part of the build process

jowilki:
Version number is currently hard coded.  This PR updates the version during the build process with the value from package.json.

Fixes https://github.com/petkaantonov/bluebird/issues/1136


petkaantonov:
The new version is not known when the files are built. I will work on a fix.

PR #1163, opened 2016-07-13
Indicate that .cast is an old alias for Promise.resolve

Laoujin:
Some old docs/tutorials etc still use `.cast`. Was a bit confusing that it was nowhere to be found in the bluebird docs.

### DIFF
--- a/docs/docs/api/promise.resolve.md
+++ b/docs/docs/api/promise.resolve.md
@@ -30,6 +30,8 @@ Promise.resolve($.get("http://www.google.com")).then(function() {
     console.log(e.statusText);
 });
 ```
+
+*For compatibility with earlier ECMAScript versions, an alias `.cast` is provided for [`Promise.resolve`](.).*
 </markdown></div>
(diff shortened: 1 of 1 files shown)


petkaantonov:
Promise.resolve is compatible with earlier ECMAScript versions, the deprecation and name change is because the standard name for this method is Promise.resolve in ES6

benjamingr:
Yeah this is just wrong. No ES version ever had `.cast` - a few drafts did but that was it.

Laoujin:
Is the .cast alias going away in bluebird? If not, the notice Could still be useful for others who are unfamiliar with the differences in es implementations of promises.

benjamingr:
There is one standard ES implementation and it does not use `.cast` @Laoujin - this has been the case for at least a few years.

petkaantonov:
Well it's not going away in the sense that we will suddenly break code, but we also don't encourage any usage of it by not including it in docs etc.

PR #1172, opened 2016-07-27
Fix Haskell's then's type

Chobbes:
The then type signature should have (a -> (Either (Promise b) b)) instead of just a -> (Either (Promise b) b). The old type signature of this function is equivalent to:

then :: Promise a -> (a -> ((Either (Promise b) b) -> Promise b))

Which isn't quite right. The type signature for (>>=) reflects my change.

Also the final b was capitalized, which is wrong since Haskell uses lower cased letters as type variables, not upper cased ones.

### DIFF
--- a/docs/docs/coming-from-other-languages.md
+++ b/docs/docs/coming-from-other-languages.md
@@ -64,7 +64,7 @@ Boost futures expose a `.then` method similar to promises and allow this functio
 A promise is a monadic construct with `.then` filling the role of `>>=` (bind). The major difference is that `.then` performs recursive assimilation which acts like a `flatMap` or a map. The type signature of `then` is quote complicated. If we omit the error argument and not throw - it's similar to:
 
 ```hs
-then::Promise a -> a -> (Either (Promise b) b) -> Promise B
+then::Promise a -> (a -> (Either (Promise b) b)) -> Promise b
 ```
 
 That is, you can return either a promise _or a plain value_ from a `then` without wrapping it.


bergus:
I guess it would also make sense to drop the `Either` in favour of `b | Promise b`, as returning a tagged union type doesn't really fit.

CbeckDrive:
@bergus as far as I am aware Either is the standard sum type in Haskell. What you are suggesting may be available in GHC 8.2 as part of this?

https://ghc.haskell.org/trac/ghc/wiki/UnpackedSumTypes

But Either is well known among Haskellers, and doesn't require knowledge of brand new GHC extensions. For documentation reflecting Haskell to Bluebird promises it seems like a good idea to go with what more people would be familiar with?

bergus:
The problem I saw with `Either` was that you would need to explictly tag it with `Left` or `Right`, which might cause some confusion. Maybe it's better to offer the Functor and the Monad type separately and simply say that it's overloaded in JavaScripts world of dynamic typing…

Chobbes:
@bergus the point of this section is to explain what the Bluebird library does / how it acts as though you are somebody coming from Haskell. Either is used for sum types in Haskell, which is what is being discussed here. This would be clear to anybody coming from Haskell.

As far as basic Haskell is concerned you are not able to have a value occupy multiple types, hence the "tagging" is required within the language. Haskell does not support subtyping. The Either type is in my opinion the best option, as the alternatives don't really make much sense from a Haskell perspective. If the purpose of this section is to help introduce people familiar with Haskell concepts to Bluebird then I don't think it makes sense to use a language which looks vaguely like Haskell, but isn't? That seems like it will breed more confusion to me.

I mean if you really want to get technical then I think you can get more accurate than (Either (Promise b) b), since this reflects just one "level" of promise, and should be something closer to a (Free Promise b), with join replaced by something similar to retract... Maybe something like:

    data Multi m a = Single a | Multi (m (Multi m a))

    joinMulti :: Monad m => Multi m a -> m a
    joinMulti (Single a) = return a
    joinMulti (Multi mma) = join $ mma >>= (return . joinMulti)

But that's way more complicated, and doesn't make the point any clearer.

Chobbes:
Bump!

I think we should make these changes because as it stands what's there is syntactically incorrect, which can make this a very confusing example!

If need be we can discuss whether or not it should use Either, or something else in another pull request, but this is a different issue.

I also have some qualms over the description stating that it acts like flatMap... Since flatMap is literaly just (>>=) on the list Monad. At least this is what I have seen in a Haskell context.

I do think it would be good to get this section right, since it is actually quite helpful if you have seen Haskell, but not Promises before.

petkaantonov:
@benjamingr @spion

benjamingr:
Let's just merge it as is it's an imporvement over the original and we can discuss more concise or accurate types later.

PR #1173, opened 2016-07-27
add api reference to the top bar

EtaiG:
I personally find the API docs very useful, and I think most people who come to bluebird will too.. it would be nice to have a link for it in the top bar :)


benjamingr:
Sounds like a good idea to me, anyone objects?

spion:
👍

benjamingr:
Thanks, will be up on the next docs rebuild

EtaiG:
Awesome! Thank you so much :) 👍

PR #1175, opened 2016-07-28
Ensure Promise.coroutine always handles yielded Promises asynchronously

### LINKED ISSUE #1170: Bluebird.coroutine() is sometimes synchronous
This is with latest Bluebird (3.4.1) and Babel (6.11.4).

Many Babel users are using `Bluebird.coroutine()` in combination with `transform-async-to-module-method` to have Bluebird shim `async/await`.

Unfortunately, it's not a perfect shim. The following code:

```js
async function foo() {
  await syncFn();
  console.log(3);
}
async function syncFn() {
  return;
}
console.log(1);
foo();
console.log(2);
```

prints `1 3 2` with `Bluebird.coroutine()` and `1 2 3` with the default `transform-async-to-generator` method.

I've created this [Phabricator ticket](https://phabricator.babeljs.io/T7511) to track but it appears the issue lies here.

The above code transpiles to this, which I have cleaned up for readability:

```js
var Bluebird = require('bluebird');

function foo() {
  return Bluebird.coroutine(function* () {
    yield syncFn();
    console.log(3);
  })();
}

function syncFn() {
  return Bluebird.coroutine(function* () {
    return;
  })();
}

console.log(1);
foo();
console.log(2);
```

This is easily runnable in any Node runtime that supports generators:

```
$ node syncCoroutines.js
1
3
2
```

Is this intended behavior?

not-an-aardvark:
fixes #1170

### DIFF
--- a/src/generators.js
+++ b/src/generators.js
@@ -175,9 +175,13 @@ PromiseSpawn.prototype._continue = function (result) {
             this._yieldedPromise = maybePromise;
             maybePromise._proxy(this, null);
         } else if (BIT_FIELD_CHECK(IS_FULFILLED)) {
-            this._promiseFulfilled(maybePromise._value());
+            Promise._async.invoke(
+                this._promiseFulfilled, this, maybePromise._value()
+            );
         } else if (BIT_FIELD_CHECK(IS_REJECTED)) {
-            this._promiseRejected(maybePromise._reason());
+            Promise._async.invoke(
+                this._promiseRejected, this, maybePromise._reason()
+            );
         } else {
             this._promiseCancelled();
         }
--- a/test/mocha/generator.js
+++ b/test/mocha/generator.js
@@ -243,6 +243,40 @@ describe("Promise.coroutine", function() {
                 assert.equal(e, error);
             });
         });
+
+        specify("when they are already fulfilled, the yielded value should be returned asynchronously", function(){
+            var value;
+
+            var promise = Promise.coroutine(function*(){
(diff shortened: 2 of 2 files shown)


spion:
Looks good, and probably isn't on a hot path since this only applies to already fulfilled/rejected promises. One alternative might be to use `async.invoke(this._promiseFulfilled, this, maybePromise._value());` instead to avoid the creation of another promise, which could be slightly more efficient. @petkaantonov ?

not-an-aardvark:
Good idea, I switched it to `async.invoke` instead.

spion:
👍

jeffijoe:
Any idea when this is coming to an npm near me? :smile:

overlookmotel:
@jeffijoe Please also see the hacky workaround I posted: https://github.com/petkaantonov/bluebird/issues/1170#issuecomment-236690921

src/generators.js
```diff
@@ -175,9 +175,13 @@ PromiseSpawn.prototype._continue = function (result) {
             this._yieldedPromise = maybePromise;
             maybePromise._proxy(this, null);
         } else if (BIT_FIELD_CHECK(IS_FULFILLED)) {
-            this._promiseFulfilled(maybePromise._value());
+            Promise._async.invoke(
```
petkaantonov: It would be better to have `async` imported directly into the generators.js module

petkaantonov:
I'll get this released today

PR #1176, opened 2016-07-29
Update value.md

antplant:
Missing word here: presumably `.isFulfilled()` only needs to be called in code paths where it's *not* already guaranteed to be fulfilled.


benjamingr:
Thanks

Issue #1181, opened 2016-08-04
No call stack with runaway promise warning


ben-page:
1) What version of bluebird is the issue happening on? **3.4.1**

2) What platform and version? (For example Node.js 0.12 or Google Chrome 32) **Node.js 6.3.1 on Windows x64**

3) Did this issue happen with earlier version of bluebird? **It appears to have started in 3.3.0**

When I receive the runaway promise warning, I never receive a call stack. 

```javascript
'use strict';
const Promise = require('bluebird');
Promise.config({
    warnings: true,
    longStackTraces: true
});

Promise.resolve()
    .then(() => {
        Promise.resolve()
            .then(() => {
                console.log('inner');
            })
    })
    .then(() => {
        console.log('outer');
    });
```

Here's the console output:
```shell
(node:3264) Warning: a promise was created in a handler but was not returned from it
inner
outer
```

It seems to be a regression in version 3.3.0, because on 3.2.2 I get a stack trace.
```
Warning: a promise was created in a handler but was not returned from it
    at Promise.resolve.then (C:\projects\test\index.js:10:17)
    at runCallback (timers.js:566:20)
    at tryOnImmediate (timers.js:546:5)
    at processImmediate [as _immediateCallback] (timers.js:525:5)
From previous event:
    at Object.<anonymous> (C:\projects\test\index.js:9:6)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
    at Function.Module._load (module.js:409:3)
    at Module.runMain (module.js:575:10)
    at run (bootstrap_node.js:352:7)
    at startup (bootstrap_node.js:144:9)
    at bootstrap_node.js:467:3

inner
outer
```

ben-page:
The call stack works with bluebird 3.4.1 on Nodejs 4.7.1. So it seems like 3.3.0 broke support for the Nodejs 5.x./6.x call stack format.

ben-page:
I've noticed a second related problem. Promise.delay() does not produce a useful call stack.

ben-page:
So the problem is that by default Nodejs 6.x does not display stack traces for warnings. There is a command line option (`--trace-warnings`) to enable them. Without this option Bluebird warnings are a lot less useful, because it can be very difficult to figure out where they are coming from.

ben-page:
[...]

PR #1183, opened 2016-08-05
fix warning stack traces in Nodejs 6.x

ben-page:
Fixes warning stack trace not being displayed in Nodejs 6.x as described in #1181


petkaantonov:
Seems really fragile (consider multiple bluebirds loaded in the process, custom warning listeners added before this code is ran etc), sorry

Issue #1059, opened 2016-03-29
Feature Request: Alternate way to suppress warnings


ben-page:
1) What version of bluebird is the issue happening on? **3.3.4**
2) What platform and version? **Node v5.9.1**
3) Did this issue happen with earlier version of bluebird? **Yes**

Library authors can't suppress the warning for runaway promises that they create, because it's not the runaway Promise that needs to resolve to null, but the outer promise. Consider the following example:

```javascript
'use strict';
var Promise = require('bluebird');
Promise.config({warnings: true, longStackTraces: true});

function otherLibrary() {
    Promise.resolve()
        .catch(function() {
            //totally recover from error
        })
        .then(function () {
            return null; //try to tell Bluebird we don't need the warning
        });
}

Promise.resolve()
    .then(function () {
        otherLibrary();
    })
    .delay(500); //on Windows node doesn't flush the console stream before exiting
```

This triggers a runaway promise warning even though, `otherLibrary` is trying to indicate that it wants to suppress the warning. So consumers of the library have to be aware of this and suppress the warning at every call, if they use promises as well.

```javascript
Promise.resolve()
    .then(function () {
        otherLibrary();
        return null; //this will suppress the warning
    })
    .delay(500); //on Windows node doesn't flush the console stream before exiting
```

I realize that the warning is thrown when the runaway promise is created, not when it's resolved. So the `otherLibrary` returning null is irrelevant. But, it would be nice to have a way to suppress the warning when creating the runaway promise.

Maybe a simple call that suppresses the warning for the next promise created.
```javascript
function otherLibrary() {
    Promise.suppressRunawayWarning();
    Promise.resolve()
        .catch(function() {
            //totally recover from error
        });
}
```

Or maybe it would be better to explicitly start a new promise chain.
```javascript
function otherLibrary() {
    Promise
        .newChain()
        .resolve()
        .catch(function () {
            //totally recover from error
        });
}
```

ben-page:
My examples are contrived to make them concise. But I'll give a specific example of why this is important.

I have a library that abstracts network communications. Internally it uses promises. But it's public interfaces uses streams and events. Non-fatal error are handled transparently. Fatal errors are caught and emitted by the error event.

The internal promises can lead to runaway promise warnings if the application uses the library also uses promises.

spion:
As far as I know warnings are now only generated if the user attempts to continue the chain after not returning a promise:

```js
Promise.resolve()
    .then(function () {
[...]

PR #1185, opened 2016-08-05
Promise.suppressRunawayWarning()

ben-page:
This is a proposed solution to the issue discussed here: #1059

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -101,6 +101,7 @@
         "IS_UNHANDLED_REJECTION_NOTIFIED": false,
         "IS_DISPOSABLE": false,
         "IS_CANCELLED": false,
+        "IS_RUNAWAY_IGNORED": false,
         "IS_CANCELLED_OR_WILL_BE_CANCELLED": false,
         "LENGTH_MASK": false,
         "LENGTH_CLEAR_MASK": false,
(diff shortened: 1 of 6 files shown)


ben-page:
The reason I didn't use done():

The existing IS_FINAL flag didn't seem to be sufficient, because it's set on the new Promise created by done() not on the runaway promise. Since I had to create a new flag, it seems more clear to create `suppressRunawayWarning()` than to muddle the affect of calling `done()`. Additionally, usage of `done()` is pretty widely discouraged online and in the docs.

petkaantonov:
Code LGTM, can you add documentation page and link in api reference?

ben-page:
@petkaantonov, sorry for the delay. I added the docs you requested and rebased on master.

spion:
Are there any test cases that test it?

chung-leong:
Any chance this will get merged? The inability to do this is leading to a lot of spurious warnings in my React.js code. The following happens fairly frequently

`
return fetch().then((value) => {
    this.setState({ foo: value });
});
`
That sets component's state variable foo and trigger a rerender. The render request can be queued or happen immediately. If it's the latter case, and one of the React lifecycle function creates a promise chain, then the warning shows up--utterly unhelpful since it's not possible to link the two chains. The fact that it doesn't always happen makes it extra frustrating. 

Personally, I'd perfer something along the line of `Promise.isolate(() => { ... })` but I guess `Promise.try(() => { ... }).suppressRunawayWarning()` is pretty obvious too.

Issue #1190, opened 2016-08-08
addEventListener("unhandledexception", ...) is ignored in workers.


lddubeau:
### Versions

Bluebird: 3.4.1. I don't know if it happens with earlier versions but I suspect it does.

Browsers: Chrome 51, Firefox 46. AFAIK, this would be reproducible on all platforms that support workers.

### Reproducing the case

Run [this plunker](https://embed.plnkr.co/m9d2ZC/), and observe the console output.

The important code is in the file `test.js`. This file is run both in the context of the main window and in a worker:

```
var inWorker = typeof window === "undefined";
var context = inWorker ? self : window;

if (inWorker) {
  self.importScripts("https://cdnjs.cloudflare.com/ajax/libs/bluebird/3.4.1/bluebird.js");
  console.log("worker running");
}

context.addEventListener("unhandledrejection", function() {
  console.log("Unahndled rejection detected by custom handler!", inWorker);
});
Promise.reject(new Error("X"));
```

### Expected behavior

The console should show these lines:

```
Unahndled rejection detected by custom handler! false
worker running
Unahndled rejection detected by custom handler! true
```

The first one unhandled rejection is from the main window. The 2nd is from the worker. Bluebird does produce warnings but those are not the issue here.

### Actual behavior

```
Unahndled rejection detected by custom handler! false
worker running
```

The unhandled rejection in the worker is never detected by the handler.

### Observations

On Chrome, removing Bluebird from the worker produces the expected behavior.

Also, assigning to `self.onunhandledrejection` works fine.

I believe the issue is with [`fireDomEvent`](https://github.com/petkaantonov/bluebird/blob/v3.4.1/src/debuggability.js#L146). `document` is not available in a worker, so the fallback is used.

`fireDomEvent` should be modified to create events using `new Event`, which is available in workers.

engineforce:
There is some error in the pull request:

`var domEvent = new CustomEvent(name.toLowerCase(), event);` resulting the detail property in the domEvent to be null.

Need to be changed to following:

`var domEvent = new CustomEvent(name.toLowerCase(), { detail: event });`

PR #1191, opened 2016-08-08
Fixes issue 1190

lextiz:


lextiz:
Fixes https://github.com/petkaantonov/bluebird/issues/1190

src/debuggability.js
```diff
@@ -143,14 +143,24 @@ Promise.hasLongStackTraces = function () {
...
+        if (typeof CustomEvent === "function") {
+            var event = new CustomEvent("CustomEvent");
+            util.global.dispatchEvent(event);
+            return function(name, event) {
+                var domEvent = new CustomEvent(name.toLowerCase(), event);
+                return !util.global.dispatchEvent(domEvent);
+            };
+        } else {
+            var event = document.createEvent("CustomEvent");
+            event.initCustomEvent("testingtheevent", false, true, {});
```
benjamingr: Why is this done here?

lextiz: This should be the case for IE and some Safari versions where CustomEvent() constructor is not supported and line 147 fails. In this case the line 155 is testing that the old-fashioned way for event initialization works. Where should it be done?

Issue #1194, opened 2016-08-09
With asCallback and finally, Unhandled Rejection is emitted.


NatalieWolfe:
1) What version of bluebird is the issue happening on?
Tested on 3.4.1, 3.3.0, and 2.10.2.

2) What platform and version? (For example Node.js 0.12 or Google Chrome 32)
Mac OS X, tested with Node 4.4.7 and 6.3.1.

3) Did this issue happen with earlier version of bluebird?
Yes, see answer to 1.

------------------
**promise.js**
```js
'use strict';

var Promise = require('bluebird');

process.on('unhandledRejection', function() {
  console.log('unhandledRejection');
})

Promise.reject('foo').asCallback(() => {
  console.log('asCallback');
}).finally(() => {
  console.log('finally');
});
```

**output**
```
$ node promise.js
asCallback
finally
unhandledRejection
```

Using the code snippet above, you'll see that `unhandledRejection` is emitted, despite the `asCallback` which will handle the rejection. I would expect the output to be just this:

**expected output**
```
$ node promise.js
asCallback
finally
```

If you remove the `.finally` bit so the promise sequence looks like this:
```js
Promise.reject('foo').asCallback(() => {
  console.log('asCallback');
});
```
Then the `unhandledRejection` event is no longer emitted:
```
$ node promise.js
asCallback
```
This seems correct to me, but is inconsistent with the `.finally` case.

ben-page:
That's because `asCallback` is designed to make a function that can accept a callback **or return a Promise**. If you continue the Promise chain from `asCallback` (with `finally`, or `then`, etc), it thinks you want the Promise version.

Look at this example:

```javascript
function CallbackOrPromise(cb) {
    return Promise.reject(new Error('Reject!'))
        .asCallback(cb);
}

//callback version
CallbackOrPromise(function(err) {
    console.log('Error From Callback: ' + err.message);
});

//promise version
CallbackOrPromise()
    .catch(function(err) {
        console.log('Error From Promise: ' + err.message);
    });
```

What you are doing is like calling `CallbackOrPromise` without a catch:
```javascript
CallbackOrPromise()
    .then(function() {
        console.log('then'); //Unhandled rejection
    });
```

NatalieWolfe:
[...]

PR #1195, opened 2016-08-11
don't return Promise from asCallback when function is paased

ben-page:
This addressed #1194. It eliminates some confusion from the usage of `asCallback`. 

And this is a very small change actually. I doesn't seem like the "both" behavior was intended. All of the unit tests still pass and the doc examples all still work.


spion:
I think this solution is the way to go. Unfortunately its a breaking change so it would have to wait for 4.0.0

I also think that throwing an error when the passed argument is not a function or undefined would make sense

benjamingr:
+1

PR #1196, opened 2016-08-12
asCallback creates a new promise and continues the chain

ben-page:
This is an alternative solution to the problem discussed in #1194. It changes `asCallback` return a new promise that resolves to the result of the callback or rejects to errors thrown in the callback.

```javascript
Promise.resolve(true)
    .asCallback(function (err, v) {
        return 'text';
    })
    .then(v => {
        console.log(v); //text
    });

Promise.resolve(true)
    .asCallback(function (err, v) {
        throw new Error('error');
    })
    .catch(err => {
        console.log(err.message); //error
    });
```

### DIFF
--- a/src/nodeify.js
+++ b/src/nodeify.js
@@ -1,62 +1,70 @@
 "use strict";
 module.exports = function(Promise) {
 var util = require("./util");
-var async = Promise._async;
 var ASSERT = require("./assert");
 var tryCatch = util.tryCatch;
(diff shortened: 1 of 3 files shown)


benjamingr:
This is fundamentally problematic since it would mean errors thrown in the callbacks are thrown in a promise context which would confuse callback users.

petkaantonov:
The alternative seems better to me

PR #1202, opened 2016-08-19
Update what-about-generators.md

LudwikJaniuk:
No, this is not a complete article, but I wish this information would have been here earlier, so that I would have discovered the promises + generators combo a month ago. Please add it while waiting for a real article so others looking here might find it and benefit from it.

### DIFF
--- a/docs/docs/what-about-generators.md
+++ b/docs/docs/what-about-generators.md
@@ -3,4 +3,5 @@ id: what-about-generators
 title: What About Generators?
 ---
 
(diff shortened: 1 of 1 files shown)


docs/docs/what-about-generators.md
```diff
@@ -3,4 +3,5 @@ id: what-about-generators
 title: What About Generators?
 ---
```
petkaantonov: Use `\[Promise.coroutine\]\(.\)` and it will automatically build up API link from it

LudwikJaniuk: Like this?

petkaantonov:
https://github.com/petkaantonov/bluebird/commit/76a6450a52a9866215655f63d1882034e215bcdf

PR #1206, opened 2016-08-24
Fix nodeify test

### LINKED ISSUE #1030: Broken test for .nodeify() return value
This is a very low priority issue, but I thought it would be worth reporting/fixing:

The [returns undefined when a callback is passed](https://github.com/petkaantonov/bluebird/blob/v3.3.3/test/mocha/nodeify.js#L63-L65) test for `.nodeify` incorrectly asserts that `.nodeify` returns `undefined` when a callback is passed.  While it actually [always returns `this`](https://github.com/petkaantonov/bluebird/blob/v3.3.3/src/nodeify.js#L60) and is [documented to do so](http://bluebirdjs.com/docs/api/ascallback.html).  The reason this test doesn't fail is that mocha does not interpret returning `false` as a failure condition and expects an assertion or promise rejection on failure.

Looking into the issue I found #151, which motivated returning `undefined`, and #168, which reverted that change for backward compatibility prior to the 2.0 release.  I assume the behavior is unlikely to change at this point, so the test can be removed or replaced with the [always returns promise for now](https://github.com/petkaantonov/bluebird/blob/v3.3.3/test/mocha/nodeify.js#L90-L99) one from the `isNodeJS` section below.

benjamingr:
Closes #1030

### DIFF
--- a/test/mocha/nodeify.js
+++ b/test/mocha/nodeify.js
@@ -61,7 +61,7 @@ describe("nodeify", function () {
     });
 
     it("returns undefined when a callback is passed", function () {
-        return 'undefined' === typeof Promise.resolve(10).nodeify(function () {});
+        return assert('undefined' === typeof Promise.resolve(10).nodeify(function () {}));
     });
 
 });


petkaantonov:
Test doesn't seem to pass?

benjamingr:
@petkaantonov right, because it fixes the test assuming https://github.com/petkaantonov/bluebird/pull/1195 is merged first.

petkaantonov:
Seems like it requires a major bump

PR #1213, opened 2016-08-27
Fixes a regression introduced by #1191 and a problem it overlooked.

lddubeau:
#1191 introduced two regressions:

1. It did not pass properly the details of the event. This was fixed in commit 288aa285fe61d4a9c360253eb5cbb66d04a81c62.

2. It did not make the event cancelable. This PR fixes this.

It also overlooked a problem with Firefox versions earlier than 48. Those versions do not make `CustomEvent` available in workers. It is a bug in Firefox but it is going to bite users of Bluebird unless Bluebird takes action to work around it. The current ESR of Firefox is version 45, which has the bug.

I've also added a test that summarily checks that events raised in workers can be caught with global handlers. Right now I have both `node tools/test`, and `node tools/test --browser` with Firefox pass all tests. 

Unfortunately, the browser tests do not pass on Chrome 51 but I've checked out v3.4.0 and v3.4.3 and found that it does not pass with these versions either. So it is not an issue I've created. (The one test I've added does actually pass on Chrome. 51.)

### DIFF
--- a/test/mocha/unhandled_rejections.js
+++ b/test/mocha/unhandled_rejections.js
@@ -677,6 +677,48 @@ if (windowDomEventSupported) {
 
         })
     });
+
+    if (typeof Worker !== "undefined") {
+        describe("dom events in a worker", function() {
+            var worker;
+            beforeEach(function () {
+                worker = new Worker("./worker.js");
+            });
+
+            afterEach(function () {
+                worker.terminate();
+            });
+
+            specify("are fired", function() {
+                var order = [];
+                return new Promise(function(resolve, reject) {
+                    worker.onmessage = function (message) {
+                        try {
+                            switch(message.data) {
+                            case "unhandledrejection":
+                                order.push(1);
+                                break;
+                            case "rejectionhandled":
+                                order.push(2);
+                                resolve();
+                                break;
+                            default:
+                                throw new Error("unexpected message: " + message);
+                            }
+                        }
+                        catch (e) {
+                            reject(e);
+                        }
+                    };
+
+                    worker.postMessage("reject");
+                }).then(function () {
+                    assert.deepEqual(order, [1, 2]);
+                }).timeout(500);
+            });
+        });
+    }
+
 }
 
 describe("Unhandled rejection when joining chains with common rejected parent", function testFunction() {
--- a/tools/browser_test_generator.js
+++ b/tools/browser_test_generator.js
@@ -33,13 +33,20 @@ module.exports = function(tests, options) {
     return Promise.join(promiseExport, main, function(promiseExport, main) {
         var browserify = require("browserify");
         var contents = promiseExport + "\n" + main + "\n" + testRequires;
-        var b = browserify({
+        var complete = browserify({
             basedir: baseDir,
             entries: stringToStream(contents)
         });
-        return Promise.promisify(b.bundle, b)().then(function(src) {
-            return writeFile(path.join(baseDir, "bundle.js"), src);
+        var worker = browserify({
+            basedir: baseDir,
+            entries: stringToStream(promiseExport),
         });
+        return Promise.join(
+            Promise.promisify(complete.bundle, complete)().then(function(src) {
+                return writeFile(path.join(baseDir, "bundle.js"), src);
+            }), Promise.promisify(worker.bundle, worker)().then(function (src) {
+                return writeFile(path.join(baseDir, "worker_bundle.js"), src);
+            }));
     }).then(function() {
         if (options.executeBrowserTests) {
             return require("./browser_test_runner.js")(options);
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ gh-pages
 node_modules
 gh-pages
 test/browser/bundle.js
+test/browser/worker_bundle.js
 js/*
 zalgo.js
 coverage/*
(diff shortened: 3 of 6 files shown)


tools/browser_test_generator.js
```diff
@@ -33,13 +33,20 @@ module.exports = function(tests, options) {
...
             entries: stringToStream(contents)
         });
-        return Promise.promisify(b.bundle, b)().then(function(src) {
-            return writeFile(path.join(baseDir, "bundle.js"), src);
+        var worker = browserify({
+            basedir: baseDir,
+            entries: stringToStream(promiseExport),
         });
+        return Promise.join(
+            Promise.promisify(complete.bundle, complete)().then(function(src) {
```
benjamingr: While refactoring - this should be `fromCallback` and you can safely use arrows since we're not in 2014 anymore.

test/mocha/unhandled_rejections.js
```diff
@@ -677,6 +677,41 @@ if (windowDomEventSupported) {
...
+                return new Promise(function(resolve, reject) {
+                    var order = [];
+                    worker.onmessage = function (message) {
+                        switch(message.data) {
+                        case "unhandledrejection":
+                            order.push(1);
+                            break;
+                        case "rejectionhandled":
+                            order.push(2);
+                            resolve();
```
benjamingr: didn't you mean to validate the order here? Also note that you might want to wrap the onmessage in a try/catch in case something throws there - reject.

test/mocha/unhandled_rejections.js
```diff
@@ -677,6 +677,41 @@ if (windowDomEventSupported) {
...
+                            break;
+                        default:
+                            throw new Error("unexpected message: " + message);
+                        }
+                    };
+
+                    worker.postMessage("reject");
+                }).timeout(500);
```
benjamingr: Can we do something smaller than 500?

benjamingr:
Left some comments but generally LGTM

tools/browser_test_generator.js
benjamingr: Nevermind, Petka mentioned that it needs to run in 0.10.x - guess we're in 2014 :D

test/mocha/unhandled_rejections.js
lddubeau: I took the timeout from the similar test done outside of workers. I supposed there was some Bluebird-specific reason for a 500 figure but I did not investigate. And I supposed that if the 500 figure in other tests did not cause problems there then it would not cause problems here either. This is a conservative stance. If I were to decide to reduce the timeout because it works for me when I try it, there's no telling whether I'd be breaking things elsewhere (e.g. if the suite is run on a platform where a timeout lower than 500 would lead to false failures).

Same reason I did not introduce arrows in this patch, by the way.

test/mocha/unhandled_rejections.js
lddubeau: Thanks. I flipped-flopped on how much I should be testing and forgot the order. I'll wrap `onmessage`, even though without wrapping Mocha will still trap any exception happening there and assign it to the correct test.

test/mocha/unhandled_rejections.js
benjamingr: Roger.

test/mocha/unhandled_rejections.js
benjamingr: Ok, this is the only thing blocking the PR at the moment.

lddubeau:
I've added the check for the order and wrapped `onmessage`.

It is unclear to me why the travis tests should be failing after the modifications I made. I'm unable to reproduce that failure here.

While performing more tests, I've discovered there are still problems with generating unhandledrejection in workers on IE and Edge. I don't know yet what the solution will be. (Preliminary observation: it seems that `dispatchEvent` in a worker causes an `InvalidStateError`.)

lddubeau:
It's just not looking good for supporting `self.addEventListener("unhandledrejection", ...` in workers across browsers.

This PR is a net gain because it fixes some of the problems and does not introduces new ones.  With this PR in place, `self.addEventListener("unhandledrejection", ...` works in:

- Chrome 51,
- Firefox 48, and 45,
- Opera 39.

However, even with the PR in place `self.addEventListener("unhandledrejection", ...` does not work in any of Microsoft's browsers, Safari or in any mobile browser I can readily test against. I don't do a lot of testing against mobile browsers. I'm sure someone could do a better job than I did here.

I kept some notes of my investigations.

In Edge 13:
 - `CustomEvent` does not exist in workers.
 - Dispatching an `Event` instance that uses a custom `type` results in an `InvalidStateError` exception.
   (`self.dispatchEvent(new Event("load"))` is okay but `self.dispatchEvent(new Event("unhandledrejection"))` fails.)

In IE11/10:
 - `CustomEvent` is undefined in workers.
 - `typeof Event === "object"`!!! I don't know what to make of this.

In Safari 9.1 (and presumably earlier), Android 4.4 (and presumably earlier, cannot test with later versions at this time),
iOS9.3 (and presumably earlier):
 - `Event` and `CustomEvent` are not functions in workers.

It seems the only cross-platform way to get notifications of unhandled rejections in workers is to use `self.onunhandledrejection`.

benjamingr:
What about `dispatchEvent`?

lddubeau:
Unsure what you mean. In all cases where `self.addEventListener("unhandledrejection", ...` does not work to catch unhandled rejections in workers, that's because `self.dispatchEvent` cannot run either because it is not possible to construct an event for it or because invoking it fails.

Issue #1189, opened 2016-08-08
api problems


frank-Z:
api promise.props . In 2nd example ,stats can be an array,but the pick method is used for object.

benjamingr:
Please elaborate

benjamingr:
Fixed in https://github.com/petkaantonov/bluebird/commit/3cc912f256d419294c833017950aeeb39e271435

PR #1216, opened 2016-08-29
Small correction to Promise.props example

luggage66:
The second example for Promise.props has a bug in how it uses lodash. It needs `_.chain()` to work, not just `_()`.

PR for #1189


benjamingr:
Thanks :)

PR #1219, opened 2016-08-31
Reject a join on no parameters.

### LINKED ISSUE #1218: `Promise.join` does not call callback if no promises provided

v2.10.2, v3.4.1

Node v6.2.1, OS X 10.9.5

Yes, consistent across v2.x and v3.x

---

(Originally raised as a side note in https://github.com/petkaantonov/bluebird/issues/1153#issuecomment-230173094)

The callback doesn't run if no promises are provided to `Promise.join()`.

```js
Promise.join( function() {
    console.log( 'callback!' );
} );
console.log('next sync statement');
```

outputs:

```
next sync statement
```

It would be pretty pointless for someone to call `Promise.join()` with no promises to join! But still you'd expect the callback to be called or an error to be thrown.

I suppose someone *might* run into this with:

```js
function foo( array ) {
    var after = function() { console.log( 'callback!' ); };
    return Promise.join.apply(Promise, array.concat( [ after ] ));
}
```

If `array` may or may not be empty, the callback being called most of the time, but skipped when the array is empty would be an unexpected result.

Of course, `Promise.all()` would be more sensible to use in this case, but you never know...

benjamingr:
Fixes #1218

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -148,3 +148,4 @@ CONSTANT(SUFFIX_NOT_IDENTIFIER, "suffix must be a valid identifier\n\n\
     See http://goo.gl/MqrFmX\n");
 CONSTANT(NO_ASYNC_SCHEDULER, "No async scheduler available\n\n\
     See http://goo.gl/MqrFmX\n");
+CONSTANT(PROMISE_JOIN_TYPE_ERROR, "Promise.join must becalled with at least one promise. It was called with only a function.");
--- a/src/join.js
+++ b/src/join.js
@@ -108,8 +108,14 @@ if (canEvaluate) {
     };
 }}
 
+function makeJoinArgumentsError() {
+    return new TypeError(PROMISE_JOIN_TYPE_ERROR);
+}
 Promise.join = function () {
     var last = arguments.length - 1;
+    if(arguments.length === 1) {
+        return Promise.reject(makeJoinArgumentsError());
+    }
     var fn;
     if (last > 0 && typeof arguments[last] === "function") {
         fn = arguments[last];
--- a/.jshintrc
+++ b/.jshintrc
@@ -95,6 +95,7 @@
         "IS_REJECTED": false,
         "WILL_BE_CANCELLED": false,
         "IS_FINAL": false,
+        "PROMISE_JOIN_TYPE_ERROR": false,
         "IS_BOUND": false,
         "IS_REJECTION_UNHANDLED": false,
         "IS_REJECTION_IGNORED": false,
--- a/test/mocha/join.js
+++ b/test/mocha/join.js
@@ -96,6 +96,16 @@ describe("Promise.join-test", function () {
 
         return a().then(assert.fail, function(e) {});
     });
+    
+    specify("gh-1218", function() {
+        function a() {
+            return Promise.join(function () {
+                
+            });
+        }
+        return a().then(assert.fail, function(e) {});
+    });
+    
 
     specify("should not pass the callback as argument, <5 arguments", function() {
         return Promise.join(1, 2, 3, function() {


src/join.js
```diff
@@ -108,8 +108,14 @@ if (canEvaluate) {
     };
 }}
 
+function makeJoinArgumentsError() {
+    return new TypeError(PROMISE_JOIN_TYPE_ERROR);
+}
 Promise.join = function () {
     var last = arguments.length - 1;
+    if(arguments.length === 1) {
```
petkaantonov: add space after `if`

benjamingr: I'll check why the lint passed. I thought it was supposed to take care of that.

src/constants.js
```diff
@@ -148,3 +148,4 @@ CONSTANT(SUFFIX_NOT_IDENTIFIER, "suffix must be a valid identifier\n\n\
     See http://goo.gl/MqrFmX\n");
 CONSTANT(NO_ASYNC_SCHEDULER, "No async scheduler available\n\n\
     See http://goo.gl/MqrFmX\n");
+CONSTANT(PROMISE_JOIN_TYPE_ERROR, "Promise.join must becalled with at least one promise. It was called with only a function.");
```
overlookmotel: Small typo here. Should be "be called" rather than "becalled".

benjamingr: Thanks!

overlookmotel:
How about the case where only a single promise and no callback is passed to `Promise.join()`?

```js
Promise.join( Promise.resolve( 1 ) ).then( function() { ... } );
```

I think the `if (arguments.length === 1)` line will create an error in this case, even though it's legal.

benjamingr:
@overlookmotel I'm modifying it to reject whenever a function is not passed as a last parameter - or if only a function is passed and no promises/values.

benjamingr:
Please take a look at the changes.

src/join.js
```diff
@@ -108,61 +108,66 @@ if (canEvaluate) {
...
-    if (last > 0 && typeof arguments[last] === "function") {
-        fn = arguments[last];
-        if (!__BROWSER__) {
-            if (last <= GENERATED_CLASS_COUNT && canEvaluate) {
-                var ret = new Promise(INTERNAL);
-                ret._captureStackTrace();
-                var HolderClass = holderClasses[last - 1];
-                var holder = new HolderClass(fn);
-                var callbacks = thenCallbacks;
+    if (arguments.length <= 1) {
```
petkaantonov: `if (last < 0)`

src/join.js
```diff
@@ -108,61 +108,66 @@ if (canEvaluate) {
...
+        return Promise.reject(makeJoinArgumentsError());
+    }
+    if(typeof arguments[last] !== "function") {
```
petkaantonov: space after if

src/join.js
```diff
@@ -108,61 +108,66 @@ if (canEvaluate) {
...
-                                              maybePromise._value(), holder);
-                        } else if (BIT_FIELD_CHECK(IS_REJECTED)) {
-                            ret._reject(maybePromise._reason());
-                        } else {
-                            ret._cancel();
-                        }
-                    } else {
-                        callbacks[i].call(ret, maybePromise, holder);
-                    }
+    var fn = arguments[last];
```
petkaantonov: Hard to see the difference when everything has changed

src/join.js
```diff
@@ -108,61 +108,66 @@ if (canEvaluate) {
...
-    if (fn) args.pop();
-    var ret = new PromiseArray(args).promise();
-    return fn !== undefined ? ret.spread(fn) : ret;
-};
-
+    args.pop();
+    var all = new PromiseArray(args).promise();
+    return all.spread(fn);
 };
+};
```
petkaantonov: line end

test/mocha/collections_thenables.js
```diff
@@ -197,47 +197,22 @@ describe("Using collection methods with thenables", function() {
...
-        return promise.then(function(v){
-            assert(v[0][0].value === 1);
-            assert(v[0][1].value === 2);
-            assert(v[0][2].value === 3);
-            assert(v[1] === 1);
-            assert(v[2] === 2);
-            assert(v[3] === 3);
-        });
-    }, 1, 2, 3);
-    testRejectSync(name, function(promise) {
```
petkaantonov: why are the tests removed

src/join.js
benjamingr: I just removed two levels of nesting because the case where a function is not passed was removed - it's actually the same code mostly - you can see it in a side by side diff.

test/mocha/collections_thenables.js
benjamingr: The tests were testing `join` in the old syntax. I added new tests for `then`ables with the new syntax. I'll add these tests again with the new syntax.

spion:
I don't think we should do this last bit - the amount of breakage is not worth it. Nothing wrong on `join` working like `all` when a function has not been passed.

petkaantonov:
Agree with spion that we should keep the old behavior and only throw if nothing but single immediate function value is passed

benjamingr:
The current behavior is confusing and causes a lot of issues for people. I absolutely think breakage is fine when we have a method that has been deprecated for many years (since before 2.0).

spion:
There is no good reason to deprecate either. `Promise.join` works just fine the way it is, baring few minor bugs.

benjamingr:
@spion it's very confusing to have a function that does two separate things for the convenience of not having to write `[]`.

spion:
It does not do separate things. The default function, if not provided is just `() => [].slice.call(arguments)`

benjamingr:
It's deprecated behavior for several years, people get confused with it all the time, it has edge cases where it doesn't work (see the issues here) and I see no reason why we should support it given it's just as easy to use `.all` which is standard.

Note that `.join` only has a performance benefit when you pass the function anyway, so this is just a function that exists in the API to avoid writing a single pair of `[]`s while being non-standard, and having a confusing case where a single function is passed as an argument.

Personally I would probably remove it altogether as well as anything that's not useful anymore with async/await and coroutines everywhere.

petkaantonov:
There is no reason to break existing code because the docs only mention the old usage at the end as side note so the risk of new code using the old syntax is small, not worth the breakage.

benjamingr:
@petkaantonov I see code that breaks in the wild no matter what modification we make to join, there is code in the wild relying on the fact that a single function argument pends forever, and code in the wild that relies on the fact `Promise.join()` is a resolved promise for an array.

I think the point of a semver major is to allow for these cleanups. 

We will always break someone's workflow



(I did a GH code search for `Promise.join`, it's in the first few pages).

petkaantonov:
If there's really code relying on function never being called we should not make that change either

benjamingr:
There is. To be clear - I'm not suggesting we merge this into 2.x, I'm suggesting we aim for 4.x, we made _much_ more breaking changes in 2.x -> 3.x breaking a lot more code and I think it was for the better. 

Stuff like "swapping the argument order of `delay`" is a lot more breakage than "in edge cases and behavior that's been deprecated for `join` bluebird will not fail fast and not do quirky things". 

It's your philosophy for the library of being genuinely useful and not correct that I'm arguing for here - in Node for example it would stay broken for life.

spion:
I still haven't seen a valid reason why this is confusing. The function takes an optional function argument, which, if not present, will place all of its other promise arguments into an array.

This is in no way confusing.

benjamingr:
```js
Promise.join(function() { //should I resolve with a promise for the function? Reject? The two cases conflict
});
```

```js
Promise.join(a, sometimesFunctionSometimesPromise); // like Array(3) vs [3], also `then`s would fire differently
```

Also, it does the same thing `.all` does for the first case. I think it was a mistake to support both cases at once.

spion:
1. Must pass at least one promise
2. Whats wrong with it? Can you explain how is it like Array(3) vs [3]

PR #1220, opened 2016-08-31
add tapError

benjamingr:
Adds `Promise.prototype.tapError` to bluebird.

The tests for the predicate version aren't passing yet (not sure why) and the doc is still missing, will hopefully follow up with those before I run out of momentum.


joepie91:
Hmm. Is this equivalent to `.error` or to `.catch`?

petkaantonov:
There should also be cancellation tests, see how .tap is tested in test/cancel.js

docs/docs/api/tapError.md
```diff
@@ -0,0 +1,128 @@
+---
+layout: api
+id: tap
```
petkaantonov: tapError

src/finally.js
```diff
@@ -3,6 +3,8 @@ module.exports = function(Promise, tryConvertToPromise) {
 var util = require("./util");
 var CancellationError = Promise.CancellationError;
 var errorObj = util.errorObj;
+var NEXT_FILTER = {};
```
petkaantonov: This needs to be the specific sentinel value that is declared in promise.js

benjamingr: Why?

benjamingr:
> Hmm. Is this meant to be equivalent to .error or to .catch?

To `.catch` (both the catch all variant and the filtered one.

joepie91:
> To `.catch` (both the catch all variant and the filtered one.

Hmm, shouldn't we name it `.tapCatch` then? Otherwise, I can see this causing some confusion for users, thinking that it's equivalent to `.error`.

src/finally.js
petkaantonov: Because otherwise it's just another object that's not special. [This](https://github.com/petkaantonov/bluebird/blob/master/src/promise.js#L549) makes it special, and this NEXT_FILTER object will fail the check,

petkaantonov:
Yeah tapCatch is probably a better name for this

src/finally.js
benjamingr: Thanks, I'll export it on `Promise` with an underscore.

benjamingr:
- Renaming to catch.
 - Fixing the doc issues.
 - I don't see `tap` in the cancellation tests. 
 - Using the same `NEXT_FILTER`.

Predicates still don't work, PTAL.

src/finally.js
```diff
@@ -103,9 +104,27 @@ Promise.prototype["finally"] = function (handler) {
...
+                                 TAP_TYPE,
+                                 undefined,
+                                 finallyHandler);
+    } else {
+        var predicate = arguments[0];
+        var handler = arguments[1];
+        return this._passThrough(handler,
+                                 TAP_TYPE,
+                                 undefined,
+                                 catchFilter(predicate, handler, this));
```
petkaantonov: it should be `catchFilter(predicate, finallyHandler, this)`

docs/docs/api/tapcatch.md
```diff
@@ -0,0 +1,128 @@
+---
+layout: api
+id: tapCatch
+title: .tapCatch
+---
+
+
+[← Back To API Reference](/docs/api-reference.html)
+<div class="api-code-section"><markdown>
+##.tapError
```
petkaantonov: s/tapError/tapCatch/g

src/finally.js
```diff
@@ -103,9 +104,27 @@ Promise.prototype["finally"] = function (handler) {
...
 };
 
+Promise.prototype.tapError = function (handlerOrPredicate) {
+    if(arguments.length === 1) {
+        return this._passThrough(handlerOrPredicate,
+                                 TAP_TYPE,
+                                 undefined,
+                                 finallyHandler);
+    } else {
+        var predicate = arguments[0];
```
petkaantonov: this needs to be typechecked  like .catch does https://github.com/petkaantonov/bluebird/blob/master/src/promise.js#L95-L99

petkaantonov:
> I don't see tap in the cancellation tests.

never mind then, remembered wrong

benjamingr:
@petkaantonov do we really need to support the multi parameter syntax? I'm fine with just supporting `.tapCatch(e => {})` and `.tapCatch(type|predicate, e => {})`

test/mocha/tapCatch.js
```diff
@@ -0,0 +1,83 @@
+"use strict";
+var assert = require("assert");
+var testUtils = require("./helpers/util.js");
+
+
+describe("tap", function () {
+    specify("passes through rejection reason", function() {
+        return Promise.reject("test").tapError(function() {
```
petkaantonov: Use new Error and assign it to variable, because this will cause a warning

test/mocha/tapCatch.js
```diff
@@ -0,0 +1,83 @@
...
+            return 3;
+        }).caught(function(value) {
+            assert.equal(value, "test");
+        });
+    });
+
+    specify("passes through reason after returned promise is fulfilled", function() {
+        var async = false;
+        return Promise.reject("test").tapError(function() {
```
petkaantonov: Use new Error and assign it to variable, because this will cause a warning

test/mocha/tapCatch.js
```diff
@@ -0,0 +1,83 @@
...
+        return Promise.resolve("test").tapError(function() {
+            called = true;
+        }).then(function(value){
+            assert(!called);
+        }, assert.fail);
+    });
+
+    specify("passes immediate rejection", function() {
+        var err = new Error();
+        return Promise.reject("test").tapError(function() {
```
petkaantonov: Use new Error because this will cause a warning

test/mocha/tapCatch.js
```diff
@@ -0,0 +1,83 @@
...
+            throw err;
+        }).tap(assert.fail).then(assert.fail, function(e) {
+            assert(err === e);
+        });
+    });
+
+    specify("passes eventual rejection", function() {
+        var err = new Error();
+        return Promise.reject("test").tapError(function() {
```
petkaantonov: Use new Error , because this will cause a warning

test/mocha/tapCatch.js
```diff
@@ -0,0 +1,83 @@
...
+                    rej(err);
+                }, 1)
+            });
+        }).tap(assert.fail).then(assert.fail, function(e) {
+            assert(err === e);
+        });
+    });
+
+    specify("passes reason", function() {
+        return Promise.reject(123).tapError(function(a) {
```
petkaantonov: Use new Error and assign it to variable, because this will cause a warning

test/mocha/tapCatch.js
```diff
@@ -0,0 +1,83 @@
...
+        }).then(assert.fail, function() { 
+            assert(called === true);
+        });
+    });
+    specify("Does not get called on predicates that don't match ", function() {
+        var called = false;
+        return Promise.reject(new TypeError).tapError(ReferneceError, function(a) {
+            called = true;
+        }).then(assert.fail, function() { 
+            assert(called === false);
```
petkaantonov: should also assert here that the error is the ReferenceError

test/mocha/tapCatch.js
```diff
@@ -0,0 +1,83 @@
...
+    specify("passes reason", function() {
+        return Promise.reject(123).tapError(function(a) {
+            assert(a === 123);
+        }).then(assert.fail, function() {});
+    });
+
+    specify("Works with predicates", function() {
+        var called = false;
+        return Promise.reject(new TypeError).tapError(TypeError, function(a) {
+            called = true;
```
petkaantonov: also assert that a is the TypeError

test/mocha/tapCatch.js
benjamingr: I got the point the first time :D

src/finally.js
benjamingr: wow, I spent like 20 minutes thinking what I'm not understanding in catchFilter and it ended up being a mistype -_-

petkaantonov:
No but the predicate should still be type checked at this point, otherwise confusing error will happen at some later point

benjamingr:
Addressed the feedback but predicate tests still don't pass.

petkaantonov:
The finallyHandler needs to be modified to:

```js
if (ret === NEXT_FILTER) {
    return ret;
} else if (ret !== undefined) {
    ...
```

src/finally.js
petkaantonov: also on top of modifying the finallyHandler, it should probably be:

```js
       return this._passThrough(catchFilter(predicate, handler, this),
                                 TAP_TYPE,
                                 undefined,
                                 finallyHandler);
```

benjamingr: That doesn't fix it either.

benjamingr:
Still no dice

petkaantonov:
Run it individually to see the reason for failure `node tools/test --run=tapCatch`, for instance you have typed RefernceError instead of ReferenceError in one of the predicate tests

benjamingr:
I did, that's not why it's failing. I see an assertionError, I don't really understand `catchFilter` very well - any chance you could take a look tonight before you release and see what I did wrong?

docs/docs/api-reference.md
```diff
@@ -53,7 +53,7 @@ redirect_from: "/docs/api/index.html"
     - [Promise.using](api/promise.using.html)
     - [.disposer](api/disposer.html)
 - [Promisification](api/promisification.html)
-    - [Promise.promisify](api/promise.promisify.html)
+    - [Promise.promisify](api/promsie.promisify.html)
```
dciccale: typo?

src/finally.js
```diff
@@ -103,9 +106,27 @@ Promise.prototype["finally"] = function (handler) {
...
+        var handler = arguments[1];
+       return this._passThrough(catchFilter(predicate, handler, this),
```
Retsam: It looks like `catchFilter` is expecting an array of predicates, not a single predicate; this is why the "Works with predicates" test is failing.

test/mocha/tapCatch.js
```diff
@@ -0,0 +1,91 @@
...
+            assert(err instanceof TypeError);
+        });
+    });
+    specify("Does not get called on predicates that don't match ", function() {
+        var called = false;
+        return Promise.reject(new TypeError).tapError(ReferneceError, function(a) {
+            called = true;
+        }).then(assert.fail, function(err) { 
+            assert(called === false);
+            assert(err instanceof ReferenceError);
```
Retsam: I'm pretty sure this should be `err instanceof TypeError`.

Retsam:
I'd love to see this feature go in; I've opened #1351 as an extension of this PR with my fixes for most of the above issues.  (Still need to add type checking for predicates)

PR #1221, opened 2016-08-31
`.map` callback called async (closes #1148)

### LINKED ISSUE #1148: `Promise.map` calls callback synchronously in bluebird v3.x

v3.0.0-3.4.1

Node v6.2.1, OS X 10.9.5

No

---

#### `Promise.map()` calling callback sync/async

The behavior of `Promise.map()` seems to have changed between bluebird v2.10.2 and v3.0.0.

In v2.x, the callback is always called asynchronously; in v3.0.0-3.4.1 the callback is called synchronously for literal values of the array.

```js
var arr = [ 1, 2, 3 ];
Promise.map( arr, function(v) {
    console.log(v);
} );
console.log('next sync statement');
```

With bluebird v2.10.2:

```
next sync statement
1
2
3
```

With bluebird v3.x:

```
1
2
3
next sync statement
```

Resolved promises in the array are also mapped over synchronously. e.g. `arr = [ 1, Promise.resolve(2), 3 ]`, gives the same result.

But any unresolved promises in the array are awaited before calling the callback on that item. This causes some puzzling behavior:

```js
var arr = [ 1, Promise.resolve(2).tap( function() {} ), 3 ];
Promise.map( arr, function(v) {
    console.log(v);
} );
console.log('next sync statement');
```

Outputs:

```
1
3
next sync statement
2
```

And also:

```js
var iterator = (function*() {
    yield 1;
    yield Promise.resolve(2).tap(function() {});
    yield 3;
})();

Promise.map( iterator, function(v) {
    console.log(v);
} );
console.log('next sync statement');
```

```
1
3
next sync statement
2
```

This behavior is the same with `Promise.filter()` too.

#### Questions

1. Was this change in behavior intended? (If so, I feel it should be noted in the docs)
2. If it wasn't intended, would this be considered Zalgo?
3. Is it inconsistent that `Promise.mapSeries()` doesn't call the callback syncronously for the first round?

overlookmotel:
Fix for #1148

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -10,6 +10,7 @@ var getDomain = Promise._getDomain;
 var util = require("./util");
 var tryCatch = util.tryCatch;
 var errorObj = util.errorObj;
+var async = Promise._async;
 
 function MappingPromiseArray(promises, fn, limit, _filter) {
     this.constructor$(promises);
@@ -22,9 +23,13 @@ function MappingPromiseArray(promises, fn, limit, _filter) {
     this._limit = limit;
     this._inFlight = 0;
     this._queue = [];
-    this._init$(undefined, RESOLVE_ARRAY);
+    async.invoke(this._asyncInit, this, undefined);
 }
 util.inherits(MappingPromiseArray, PromiseArray);
+
+MappingPromiseArray.prototype._asyncInit = function() {
+    this._init$(undefined, RESOLVE_ARRAY);
+};
 
 // The following hack is required because the super constructor
 // might call promiseFulfilled before this.callback = fn is set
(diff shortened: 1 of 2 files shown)


petkaantonov:
I wonder if there is same issue with other PromiseArray subclasses as well? like reduce.js

benjamingr:
Changes LGTM, if other promise array classes do this - we should just do it in init

overlookmotel:
All other methods call the callback async. I've tested them all. It's only `map` and `filter` which have this problem, and this PR fixes both.

overlookmotel:
NB see the tests here: https://github.com/overlookmotel/cls-bluebird/tree/refactor

overlookmotel:
Any chance of releasing to npm please?

petkaantonov:
I'll be able to release tonight

overlookmotel:
@petkaantonov OK brilliant. Thank you.

petkaantonov:
released

PR #1227, opened 2016-09-03
Quick start example on the "Getting started" page

jesperlindstrom:
Took me 5 minutes to figure out how to do the simplest thing – to return a promise and act upon it. I had to dig through the "API reference" and look for things that seemed appropriate. 

The "Getting started" guide only mentions installation, but would definitely benefit from a quick start example. The API reference is definitely good for more detailed information, but the "quick start" should suffice for lots of new users who just want to use promises in their own methods.

### DIFF
--- a/docs/docs/getting-started.md
+++ b/docs/docs/getting-started.md
@@ -35,22 +35,32 @@ When using script tags the global variables `Promise` and `P` (alias for `Promis
 
 ## Quick start
 
-How to define a function that returns a promise.
+An example function that performs an AJAX request and returns a promise.
 ```javascript
-function doSomething() {
+function doGet() {
   return new Promise(function(resolve, reject) {
-    // Do asynchronous things here, such as an AJAX request or a database look up
-    var result = "Hello world";
+    var xhr = new XMLHttpRequest();
     
-    return resolve(result);
+    xhr.addEventListener("load", function() {
+      resolve(xhr);
+    }, false);
+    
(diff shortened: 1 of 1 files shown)


petkaantonov:
I'd prefer something more realistic and practically useful, e.g. converting some existing API to promises (via promisifyAll) and using the new promisified api

petkaantonov:
@benjamingr since you answer a lot of noob stack overflow questions

jesperlindstrom:
Thanks for your reply. I was just assuming the "primary use" out of my own experience, as a beginner who just wanted to use promises in his own methods. Maybe this isn't the main use case with Bluebird, as you say.

petkaantonov:
Yea in the code comment you mention that you should do something asynchronous. Maybe it could be a fuller implementation of promisified XHR e.g.:

```js
 function doGet(url) {
   return new Promise(function(resolve, reject) {
    var xhr = new XMLHttpRequest();
    xhr.addEventListener("load", function() {
      resolve(xhr);
    }, false);
    xhr.addEventListener("error", function(ev) {
      resolve(ev.error);
    });
    xhr.open("GET", url);
    xhr.send(null);
   });
 }
```

jesperlindstrom:
I updated the example with your suggestion. Again, this is mainly relevant if you feel like this is the primary use case with Bluebird. I'm merely assuming that beginners are looking for this, out of my own experience.

PR #1239, opened 2016-09-19
add isCancelled() regression tests

finnigantime:
Adding some regression tests around isCancelled(). These tests are green against bluebird master:

screen shot 2016-09-19 at 8 09 39 am

In our fork of bluebird, we moved over from bluebird2 to bluebird3 a few months ago so we're on a slightly older version of bluebird still. We have some issues with this version around cancellation, and the second unit test here isolates one of these issues - it fails in our fork:

screen shot 2016-09-19 at 8 10 10 am

Looks like we can merge the latest release of bluebird to fix the issue on our side. Adding these tests to protect this behavior.

### DIFF
--- a/test/mocha/cancel.js
+++ b/test/mocha/cancel.js
@@ -281,7 +281,7 @@ describe("Cancellation", function() {
         });
     });
 
-    specify("Can be used for breaking chains early", function() {
+    specify("can be used for breaking chains early", function() {
         var called = false;
         var p = Promise.resolve(1)
             .then(function(data) {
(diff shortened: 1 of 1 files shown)


test/mocha/cancel.js
```diff
@@ -709,7 +709,21 @@ describe("Cancellation", function() {
         req.cancel();
         var resolve;
         return new Promise(function(_, __, onCancel) {resolve = arguments[0]});
-    })
+    });
+
+    specify("isCancelled() synchronously returns true after calling cancel() on pending promise", function() {
+        var resolver = Promise.pending();
```
benjamingr: Let's not use `Promise.pending` in new tests?

test/mocha/cancel.js
```diff
@@ -709,7 +709,21 @@ describe("Cancellation", function() {
...
+        resolver.promise.cancel();
+        assert(resolver.promise.isCancelled());
+    });
+
+    specify("isCancelled() synchronously returns true after calling cancel() on promise created from .then()", function() {
+        var resolver = Promise.pending();
```
benjamingr: Same, let's not use Promise.pending in new tests

test/mocha/cancel.js
finnigantime: I'm happy to clean this up. What should it be replaced with?

finnigantime:
Feedback addressed.

finnigantime:
@benjamingr Does this look ok to merge?

finnigantime:
@benjamingr @petkaantonov ping :)

Issue #590, opened 2015-04-17
Can't find Promise.cast() method in this DOCs


PutinHuilo:
Something wrong with that method?

spion:
Promise.cast() has been deprecated for a long time now, ever since ES6 decided to drop it. You should switch to `Promise.resolve()` for all your casting needs

Laoujin:
Could I add this to: http://bluebirdjs.com/docs/deprecated-apis.html?

There is still documentation, tutorials etc out there that use `.cast` which makes things complicated for those who do not have this context (like me, 5 minutes ago)

The `Promise.cast` function is still an alias in the latest code base aswell
https://github.com/petkaantonov/bluebird/blob/master/src/promise.js#L191

PR #1250, opened 2016-09-30
Deprecate message for Promise.cast

dciccale:
Promise.cast has been deprecated as mentioned here https://github.com/petkaantonov/bluebird/issues/590


dciccale:
no activity, closing

PR #1256, opened 2016-10-08
Add Swift language to coming-from-other-languages

AndrewBarba:
Add Swift language to coming-from-other-languages


benjamingr:
This is not a "coming from", this is a link to your library :) add some coming from.

AndrewBarba:
Hey you have to at least appreciate the effort :)

benjamingr:
Are you going to work on this and add some content :)?

AndrewBarba:
Yes! There's no less content here than the Objc section, but I want to wait for more feedback on the library so I can comment on things people actually like/use. It would make sense to bring up the features people are actually familiar with after all

benjamingr:
Good luck with your library

PR #1262, opened 2016-10-14
feat(name-translator): allow custom naming fn for promisifyAll

### LINKED ISSUE #1261: [feature-request] promisifyAll allow full control over promisified method name
v3.4.6

node.js 6.5.x

yes

It would be useful to have more control over the generated name when using `promisifyAll`. Currently we are allowed to specify a suffix, however in our project we wanted to use `promisifyAll` to mutate methods exposed by a native module in the form `_method` such that the promisified version would be `method`. As far as I'm looking at the code, this isn't possible but should be relatively easy to add - if you're amenable I could submit a PR.

mbroadst:
Allows users fine-grained control over the name of a method that
is being promisified with promisifyAll.

closes #1261

### DIFF
--- a/src/promisify.js
+++ b/src/promisify.js
@@ -245,7 +245,7 @@ var makeNodePromisified = canEvaluate
     ? makeNodePromisifiedEval
     : makeNodePromisifiedClosure;
 
-function promisifyAll(obj, suffix, filter, promisifier, multiArgs) {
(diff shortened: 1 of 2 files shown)


mbroadst:
totally open to bikeshedding on naming here

mbroadst:
@benjamingr @petkaantonov any comments on this?

petkaantonov:
What about cases where the translator causes name conflict?

PR #1267, opened 2016-10-20
Promise.reject() doesn't cause OperationalError wrapping since 3.0

cdhowie:
Update OperationalError docs, which documented pre-3.0 behavior.

### DIFF
--- a/docs/docs/api/operationalerror.md
+++ b/docs/docs/api/operationalerror.md
@@ -14,8 +14,7 @@ new OperationalError(String message) -> OperationalError
 ```
 
 
-Represents an error is an explicit promise rejection as opposed to a thrown error. For example, if an error is errbacked by a callback API promisified through undefined or undefined
-and is not a typed error, it will be converted to a `OperationalError` which has the original error in the `.cause` property.
+Represents an error is the result of a node-style errback from a promisified API call. For example, if an error is errbacked by a callback API promisified through undefined or undefined and is not a typed error, it will be converted to a `OperationalError` which has the original error in the `.cause` property.
 
 `OperationalError`s are caught in [`.error`](.) handlers.
 </markdown></div>


benjamingr:
I use OperationalError and subclasses thereof in my own APIs all the time having to do nothing with promisification.

cdhowie:
@benjamingr I think you're missing the point.  Of course you can continue to explicitly use OperationalError.

In Bluebird 2.x, returning `Q.reject(foo)` from a handler would cause the new promise to be rejected with an OperationalError having its "cause" property set to the foo object that was rejected with.  In 3.0 this is no longer the case; the new promise is rejected with the foo object as the reason directly.

That's the only thing my change tries to address; the docs were explaining the 2.x behavior, not the 3.0 behavior.

benjamingr:
I'm not missing the point - while this change is _technically_ correct it creates the misconception of where `.error` should be used. 

> +Represents an error is the result of a node-style errback from a promisified API call.

Is not a very useful indicator

cdhowie:
> Is not a very useful indicator

To my knowledge, that's the only (or at least the most common) event in which _Bluebird itself_ will create an OperationalError.

The documentation will at least be more accurate than it was before this PR.  Don't let the perfect be the enemy of the good -- or at least give some feedback that I can use to help improve the PR.  Right now your comments amount to "the new text isn't 100% correct."  Okay, but it's closer to correct than it was before.  If you want it to be 100% correct then you're going to have to give me better criticism.

spion:
FWIW my deliberation is because of this article:

https://www.joyent.com/node-js/production/design/errors

Ideally we would link to it since OperationalErrors are based on that.

However, I have some doubt now that the entire premise of the article is not correct. There aren't two types of errors. Instead, there are two ways for users to consume an API: one is assuming an error cannot happen (e.g. due to external knowledge, for example, that the index cannot be out of range etc) and another is assuming that errors can happen. So even though the producer generates the error and determines its type, only the consumer knows whether the error is expected...

So I think this PR is actually an accurate description of the current state of things...

cdhowie:
The original wording of the docs seemed to imply, if not outright state, that node-style errbacks are only wrapped in an OperationalError if the error value is not a proper Error object.  I've just verified that, at least in node, _all_ errbacks from node-style calls (`Promise.fromCallback`) result in an OperationalError wrapping the errbacked error object.

That's rather unfortunate; we're going to have to wrap `fromCallback` in our code to catch OperationalErrors, extract the original error, and reject with it.  I'm not sure how much performance we will lose doing this, but at least it will be compatible with Q's behavior.

```js
let fromCallback = Promise.fromCallback.bind(Promise);

Promise.fromCallback = () =>
    fromCallback.apply(void 0, arguments)
    .error(e => Promise.reject(e.cause || e));
```

Risking getting a bit off-topic, I can't fathom the thought process that went into the decision to wrap all errors like this.  Just because an error object comes out of a different asynchronous API, it's treated as second-class within Bluebird.  That's silly and complicates error handling (for us) in many cases.

Hopefully this will be rectified in 4.0, or maybe we can at least get a config option to disable it.  In the meantime, at least the docs can document what really happens in the 3.x line.

cdhowie:
On October 20, 2016 6:25:31 PM EDT, Benjamin Gruenbaum <notifications@github.com> wrote:
>I use OperationalError and subclasses thereof in my own APIs all the
>time having to do nothing with promisification.

I think you're missing the point.  Of course you can continue to explicitly use OperationalError.

In Bluebird 2.x, returning Q.reject(foo) from a handler would cause the new promise to be rejected with an OperationalError having its "cause" property set to the foo object that was rejected with.  In 3.0 this is no longer the case; the new promise is rejected with the foo object as the reason directly.

That's the only thing my change tries to address; the docs were explaining the 2.x behavior, not the 3.0 behavior.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #1275, opened 2016-11-01
Added stack detection for Edge.

thcgit:
Changed captureStackTrace so that it detects the MS Edge browser for long stack trace support.

Couldn't find unit tests for the other browser checks so I didn't add one.

### DIFF
--- a/src/debuggability.js
+++ b/src/debuggability.js
@@ -860,6 +860,20 @@ var captureStackTrace = (function stackDetection() {
         };
     }
 
+    //Edge
+    if (typeof err.stack === "string" &&
+        typeof Error.stackTraceLimit === "number" &&
+        v8stackFramePattern.test(err.stack.split("\n")[1])) {
+        stackFramePattern = v8stackFramePattern;
+        formatStack = v8stackFormatter;
+        return function captureStackTrace(o)
+        {
+            Error.stackTraceLimit += 6;
+            o.stack = new Error().stack;
+            Error.stackTraceLimit -= 6;
+        };
+    }
+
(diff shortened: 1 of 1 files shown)


benjamingr:
I confirm this is working for me. Thanks!

@petkaantonov PTAL and let me know if you have any reservations about merging.

petkaantonov:
It looks very similar to v8, maybe they can be put together for less code?

thcgit:
Hmm, we could merge that block into something like this:

```
if (typeof Error.stackTraceLimit === "number") {
    var captureStackTrace;
    if (typeof Error.captureStackTrace === "function") {
        Error.stackTraceLimit += 6;
        captureStackTrace = Error.captureStackTrace;
        shouldIgnore = function(line) {
            return bluebirdFramePattern.test(line);
        };
    } else if (typeof err.stack === "string" &&
        v8stackFramePattern.test(err.stack.split("\n")[1])) {
        captureStackTrace = function (o) {
            o.stack = new Error().stack;
        };
    }
    if (captureStackTrace) {
        stackFramePattern = v8stackFramePattern;
        formatStack = v8stackFormatter;
        return function(receiver, ignoreUntil) {
            Error.stackTraceLimit += 6;
            captureStackTrace(receiver, ignoreUntil);
            Error.stackTraceLimit -= 6;
        };
    }
}
```

But would either have to move the creation of err to before this code, or do new Error().stack in the condition where we test if stack is a string.

petkaantonov:
Yes, using less code to implement the same thing is preferable

Issue #1280, opened 2016-11-14
longStackTraces lose the context for SyntaxErrors


broofa:
1) What version of bluebird is the issue happening on? **v3.4.6**

2) What platform and version? (For example Node.js 0.12 or Google Chrome 32)  **MacOSX, node v6.9.1**

It appears Bluebird loses the file/lineno info on longStackTraces when a SyntaxError is raised inside a promise.  For example ...

With `longStackTraces: false`, Syntax Errors [correctly] appears like this:

```
boss/db/Command.js:114
  }
  ^
  SyntaxError: missing ) after argument list
  at Object.exports.runInThisContext (vm.js:76:16)
  at Module._compile (module.js:542:28)
  at Object.Module._extensions..js (module.js:579:10)
(etc...)
```

With `longStackTraces: true`, that same error appears like this:
```
SyntaxError: missing ) after argument list
  at Object.exports.runInThisContext (vm.js:76:16)
  at Module._compile (module.js:542:28)
  at Object.Module._extensions..js (module.js:579:10)```

**Note**: the first 3 lines - the lines with the context for the SyntaxError - are missing.   As you can imagine, this makes diagnosing syntax errors difficult.  I find myself having to restart my app with the appropriate longStackTraces setting, depending on the type of error. :-/

broofa:
Any ETA on when this might be merged/published?  This is a non-trivial pain point for me - any time I have a syntax error in my code I have to disable long stack traces to determine the location.  :-(

petkaantonov:
I'll try to get release today

SimonSchick:
@petkaantonov can this be closed with https://github.com/petkaantonov/bluebird/releases/tag/v3.4.7 ?

benjamingr:
Yes.

PR #1281, opened 2016-11-14
Fix for #1280 - Don't drop SyntaxError context for longStackTraces

broofa:
SyntaxErrors in node.js are different from other Errors in JS - the first 3 lines of the stack specify the file, lineno, and code snippet where the error occurred.  Previously, longStackTraces were stripping this information, which made identifying the fix problematic.  This diff keeps this info in the stack.


benjamingr:
I'm +1 on this change.

spion:
Also 👍

broofa:
Sigh... futzing around with this to get a branch I can make the necessary changes needed to `npm install` from my github repo and have made a clusterfuck of things.  Closing and will re-request shortly.

PR #1282, opened 2016-11-16
Fix bad indentation in mutation observer scheduler

timotheeg:
I assume bad indentation was not intentional, the indentation under the `if (toggleScheduled)` made the code rather confusing :)


benjamingr:
LGTM

PR #1285, opened 2016-11-18
Delay Error.stack until the last possible moment

za-creature:
Accessing Error.stack is about 10 times slower than creating a new Error in vanilla Node.JS. This is made orders of magnitude worse when using coffeescript or a stack beautification lib that further changes the trace. As such, it should only ever be used when needed. The current implementation of `Promise.coroutine` obtains the stack whenever a generator is converted to a promise and while this is a one-time cost per promisified generator, if called from within a closure this needlessly obtains the stack when everything is working as expected.

This changes that behavior to only call .stack when the returned promise will throw. While benchmarking, I observed that accessing the .stack property appears to have a similar performance profile with a cached function, so local caching shouldn't be needed. It might make sense to cache the extra .split().slice().join() chain, but I believe that should be a separate PR.

Benchmark: https://gist.github.com/za-creature/28ca43947ee2c0a7ead33558ac4a9fa1
Results (Node.JS 4.6.1):
```
Different error w/o stack 484
Different error w/ stack 4809
Different error w/ stack w/o reuse 5241
Same error w/ stack w/ plain reuse 4886
Same error w/ stack w/ cached reuse 5209
```

### DIFF
--- a/src/generators.js
+++ b/src/generators.js
@@ -30,7 +30,7 @@ function promiseFromYieldHandler(value, yieldHandlers, traceParent) {
     return null;
 }
 
-function PromiseSpawn(generatorFunction, receiver, yieldHandler, stack) {
+function PromiseSpawn(generatorFunction, receiver, yieldHandler, context) {
     if (debug.cancellation()) {
         var internal = new Promise(INTERNAL);
         var _finallyPromise = this._finallyPromise = new Promise(INTERNAL);
@@ -43,7 +43,7 @@ function PromiseSpawn(generatorFunction, receiver, yieldHandler, stack) {
         var promise = this._promise = new Promise(INTERNAL);
         promise._captureStackTrace();
     }
-    this._stack = stack;
+    this._context = context;
     this._generatorFunction = generatorFunction;
     this._receiver = receiver;
     this._generator = undefined;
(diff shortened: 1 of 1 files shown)


spion:
I believe this used to be implemented like you are saying, but that resulted with memory leaks with infinite promise chains, and now bluebird converts the stacks to strings once the chains reach a certain length...

Afaict the code you modified is only executed during program load when the `Promise.coroutine` generator function wrapper is called, so it doesn't have an effect at run time...

za-creature:
Sorry, I should've been more clear with the whole 'called from a closure'.

While unfortunate, our current usage pattern involves calling `Promise.coroutine` dynamically. We are in the process of coroutinifying all of our codebase, but until we finish that process, we are using a helper to add coroutines into existing async code:

```js
async.auto({
  "something": coroutine.auto(function* () {
    value = yield Db.query(...)
    return value
  }),
  "something_else": ["something", ...]
})
```

Where `coroutine.auto` calls `Promise.coroutine` and the `async.auto` block is part of a route handler (so gets executed on every req).

---

As for the memory leak, I'm not sure I understand this correctly, but it seems to me like the stack never leaves the local context un-stringified. This PR just delays that process until it actually _needs_ the stack. If storing errors as properties creates a leak (maybe they're not GC'd properly), I could implement a local function that explicitly `delete`s the error object once the stack was generated and cached.

benjamingr:
If you take the `Promise.coroutine` call outside of that function and call it once-per-function the problem goes away.

spion:
Oh, now I understand what you're trying to do.

Still, if you use coroutine in that way, the unserialized stack traces may cause memory leaks. See https://github.com/kriskowal/q/issues/111

petkaantonov:
Coroutine is assumed to be called once per program startup per function, so it is slow.

PR #1312, opened 2016-12-31
Correct a few found typos in async-dialogs.md

sbauer322:
Nothing too interesting here, just some discovered typos.


benjamingr:
Ty

PR #1315, opened 2017-01-04
Fixed #1314:  yielding a function should not call the function

### LINKED ISSUE #1314: Promise.coroutine: yielding a function should not call the function
  3.4.7

  Tested on Node 7, but I'm guessing it's true for all platforms

  Probably.

Given:
```js
(Promise.coroutine(function*() {
  yield function () { console.log('hi'); };
}))();
```
You would not expect "hi" to be written to log, because we're yielding a _value_ (which is a function), and not calling the function.

But it does print "hi". And then returns a rejected error.

I have prepared a pull-request that contains a fix, which I will submit in a few minutes.

giltayar:
When rejecting a non-promise value, string.replace is used, and the value is passed as second parameter. But string.replace, if the second parameter is a function, will call that function.
This commit fixes this by using value.toString().

### DIFF
--- a/src/generators.js
+++ b/src/generators.js
@@ -160,8 +160,7 @@ PromiseSpawn.prototype._continue = function (result) {
             if (maybePromise === null) {
                 this._promiseRejected(
                     new TypeError(
-                        YIELDED_NON_PROMISE_ERROR.replace("%s", 
-                            value == null ? value : value.toString()) +
+                        YIELDED_NON_PROMISE_ERROR.replace("%s", String(value)) +
                         FROM_COROUTINE_CREATED_AT +
                         this._stack.split("\n").slice(1, -7).join("\n")
                     )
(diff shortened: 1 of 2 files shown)


src/generators.js
```diff
@@ -160,7 +160,8 @@ PromiseSpawn.prototype._continue = function (result) {
             if (maybePromise === null) {
                 this._promiseRejected(
                     new TypeError(
-                        YIELDED_NON_PROMISE_ERROR.replace("%s", value) +
+                        YIELDED_NON_PROMISE_ERROR.replace("%s", 
+                            value == null ? value : value.toString()) +
```
petkaantonov: `value+ ""` is sufficient

benjamingr:
LGTM, @petkaantonov ?

giltayar:
Funny what I just saw in /r/javascript: https://www.reddit.com/r/javascript/comments/5m2e8j/string_vs_tostring/dc0wzr5/

benjamingr:
Thanks!

PR #1317, opened 2017-01-13
Update promise.each.md

charlie-s:
I fixed a few grammatical issues and moved the sentence "If the iterator function returns a promise..." to the first paragraph where it seems to belong since it's part of the iterator logic. It's a pretty important thing to keep in mind and it seemed like an afterthought.

I also posted a few examples in the comments at http://bluebirdjs.com/docs/api/promise.each.html of using `.each`. I think these examples are helpful because people seem to think that passing an array of promises to `.each` means the promises haven't executed yet, which is not the case, and the examples that I wrote demonstrate how to achieve just that. Are there guidelines for writing examples? Can they be further refined and included in the docs on this page?

### DIFF
--- a/docs/docs/api/promise.each.md
+++ b/docs/docs/api/promise.each.md
@@ -19,10 +19,9 @@ Promise.each(
 
 [api/promise.each](unfinished-article)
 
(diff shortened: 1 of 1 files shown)


benjamingr:
Thanks!

PR #1318, opened 2017-01-13
Added dedicated webpack entry point.

worldsense-tms:
I'm using Bluebird 3.4.7 with webpack 1 and noticed it was clobbering the global variable `P`. I could see the snippet at the end of the script that it puts itself in the global namespace.

```js
                       ;if (typeof window !== 'undefined' && window !== null) {
                             window.P = window.Promise;
                        } else if (typeof self !== 'undefined' && self !== null) {
                             self.P = self.Promise;
                        }
```

I investigated a bit and found out that webpack will use the "browser" field from package.json if it's available. From [the webpack docs](http://webpack.github.io/docs/configuration.html#resolve-packagemains):

> Check these fields in the package.json for suitable files.
> Default: ["webpack", "browser", "web", "browserify", ["jam", "main"], "main"]

This means that webpack picks the "browser" version instead of "main". Running webpack --display-modules confirmed it:

```
[5] ./~/bluebird/js/browser/bluebird.js 178 kB {0} [built]
```

Since Blubird's [installation docs](http://bluebirdjs.com/docs/install.html#browserify-and-webpack) do not mention it, I think it's expected to include `release/bluebird.js` from the `main` field.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
   },
   "readmeFilename": "README.md",
   "main": "./js/release/bluebird.js",
+  "webpack": "./js/release/bluebird.js",
   "browser": "./js/browser/bluebird.js",
   "files": [
     "js/browser",


worldsense-tms:
I was able to work around this with an alias in webpack.config.js:

```js
alias: {
  'bluebird': 'bluebird/js/release/bluebird'
}
```

Perhaps you'd like to change the [installation docs](http://bluebirdjs.com/docs/install.html#browserify-and-webpack) instead?

le0nik:
For those confused why it doesn't work in webpack 2, default mainFields are different in it

```
mainFields: ["browser", "module", "main"]
```

In order to make this work you need to specify them manually

```
mainFields: ["webpack", "browser", "module", "main"]
```

An alternative(which I was using prior to this PR) would be to use an alias for bluebird:

```
resolve: {
  alias: {
    bluebird: path.join(__dirname, './node_modules/bluebird/js/release/bluebird.js'),
  }
}
```

P.S.: Now you can also decrease bundle size by stripping unneeded modules from it like so:

```js
    new webpack.NormalModuleReplacementPlugin(  /\/node_modules\/bluebird\/js\/release\/(?:assert|each|settle|using|any|race|call_get|filter|generators|nodeify|promisify|map|reduce|settle|some|progress)(?:\.js)?$/,
      path.join(__dirname, './node_modules/lodash/noop.js')
    ),

```

PR #1329, opened 2017-01-25
Added additional keywords to package.json

bkim49:
Hello, I have added additional keywords to the package.json. 
Let me know if there are any other fields you wanted added/removed.


benjamingr:
bluebird is already found as the package name, browserstack is unrelated and most things on NPM are libraries anyway.

PR #1331, opened 2017-01-26
Fix using setting longStackTraces through Promise.config function

vichugunov:


petkaantonov:
Can you explain this code? It is ignoring the passed in longStackTraces option and using whatever is the default from environment variables...

vichugunov:
Sure. I noticed, that passed through opts __longStackTraces__ opts is ignored and not assigned to __config__ within __Promise.config__ function. 

I mean, that executing the following:
```js
Promise.config( {
                    longStackTraces: true
                } );
```
actually do not set __longStackTraces__ into __config__ variable without that patch, and when __longStackTraces__ called firstly, __config.longStackTraces__ equals to false, that's not ok.

Just take a look at the following snippet, that you already have below within your code [link to line](https://github.com/petkaantonov/bluebird/blob/master/src/debuggability.js#L264):

```js
 if ("warnings" in opts) {
        var warningsOption = opts.warnings;
        config.warnings = !!warningsOption;
....
}
```

You see, that __config.warnings__ is set, accordingly to __opts.warning__? 
So, I've done the same for __opts.longStackTraces__, cause it's didn't set to __config.longStackTraces__

vichugunov:
Hm, as for me, it's strange to give an ability to config property, that actually cannot be configurable.
Property __config.longStackTraces__ is set only [here](https://github.com/petkaantonov/bluebird/blob/master/src/debuggability.js#L126) and [here] (https://github.com/petkaantonov/bluebird/blob/master/src/debuggability.js#L135)

Consequently, it is set regardless options, passed to __Promise.config__.
So, it can be two ways to resolve it:
* add it to __config.longStackTraces__ , accordingly to passed corresponding option
* remove this option from config and do not give ability to set it through __Promise.config__

Am I right?

vichugunov:
That fix was fit for me when working within safari on my mac.
Without that I got an error.
I will try to reproduce it with more details and write some tests.

vichugunov:
These errors are exposes within IE edge 11 & Safari on mac. 
It seems, that ___queue__ filled with some promises before __Promise.config__. It seems strange, that within ie 11 & chrome it doesn't reproduce

PR #1332, opened 2017-01-26
check if ms parameter is a number for delay method

jbpionnier:


fclaeys:
👍

benjamingr:
Bluebird used to do this but it was removed because users expected `setTimeout` like behavior and things having `.valueOf` working.

You can use a type checker to enforce this if you'd like.

jbpionnier:
It's a shame, it can cause an edge effect... 🤕

PR #1351, opened 2017-02-27
Fixes to tapError implementation/tests

Retsam:
Amends #1220


analytik:
Not sure if this is the right place, but tapCatch documentation is missing: http://bluebirdjs.com/docs/api/tapCatch.html (gives me a github pages 404).

Retsam:
@analytik Good catch, looks like the link is wrong.  Opened #1354.

PR #1358, opened 2017-03-13
Update warning-explanations

ry8806:
made slight corrections to the use of English in the text.

IMO it reads slightly better, happy to discuss though :smile:


benjamingr:
Thanks

PR #1360, opened 2017-03-19
Update then.md

ppjerry:
updated the argument explanation of both fulfilled and rejected callback functions


benjamingr:
I don't think this explains things better.

PR #1362, opened 2017-03-20
Fix/readme badges

thechutrain:
the readme.md badges were not properly displaying.

https://github.com/thechutrain/bluebird/tree/fix/readme-badges


benjamingr:
Thanks

Issue #1346, opened 2017-02-23
There should be a way to generate warnings when reject is called after resolve


aalexgabi:
I have lost many hours investigating a problem that arrived due to a promise being resolved and rejected afterward. Here is an example to illustrate the problem I had:
```javascript
var Promise = require('bluebird')

new Promise(function (resolve, reject) {
    console.log('resolving')
    resolve()

    setTimeout(function () {
        console.log('rejecting')
        reject(new Error())
    }, 0)
}).then(function () {
    console.log('resolved')
}).catch(function (err) {
    console.log('rejected', err)
})
```

I tried to activate warnings but I din't get any:
```
root@874aba35616a:/tmp# BLUEBIRD_WARNINGS=1 node example.js 
resolving
rejecting
resolved
root@874aba35616a:/tmp# 
```

When I don't supply an error to reject, I get a warning so I'm confident that the warning mechanism gets activated using the environment variable.

Is there a way to have these warnings? If not can you add this feature? I bet it would save a lot of debugging time.

spion:
`new Promise` should be used very sparingly (almost never in fact). Can you post some realistic examples where this happens?

benjamingr:
I actually use this fact as a feature (resolving after rejecting)

aalexgabi:
Here is the code I was debugging:
```js
return new Promise(function (resolve, reject) {
        var csvStream = csv.parse({ delimiter: self.request.separator || ',', columns: true, relax_column_count: true });
        csvStream.on('pipe', function (src) {
            src.on('data', function (data) { self.progressSize += data.length; });
        });
        self.sftpConnection.readAndPipeFile(file.filename, csvStream).catch(function(err){
            err.context = err.context || {};
            err.context.filename = file.filename
            reject(err);
        });
       // BulkPromiseStream is a transformer that waits for promise returned by the mapper to fulfill before acknowledging the fake "write" and emits error on rejection
        csvStream.pipe(new streamTools.BulkPromiseStream(function(slice) {
            return self.processSomeRows(slice).then(function (sliceReport) {
                addReports(report, sliceReport);
                lines += slice.length;
            }).then(function () {
                return self.updateProgress({
                    totalFiles: totalFiles,
                    processedFiles: fileNumber,
                    totalSize: self.totalSize,
                    processedSize: self.progressSize,
                    currentProcessedLines: lines,
                    currentReport: JSON.stringify(report),
                });
            });
        }))
[...]

PR #1363, opened 2017-03-22
Issue 1346 - Warning on reject after resolve.

gregory-bramwell-zefr:
#1346 Warning on reject after resolve.


gregory-bramwell-zefr:
As per issue discussion, going to close out this pull request.

PR #1366, opened 2017-04-04
add Aigle to benchmark

suguru03:
I have implemented [`Aigle`](https://github.com/suguru03/aigle) library. 
It is inspired by `Bluebird` and also it is taking another approach not to slow down in any cases.
Could you possibly add it to your benchmark?


KoteKotlyarov:
@suguru03 any updates on your lib ? I have read you article from 2017 and since then was unable to find any performance benchmarks for your lib. 4 years have passed, is here any improvements ? Are your benchmarks actual ?

PR #1371, opened 2017-04-13
add note about simultaneous behaviour

timaschew:


benjamingr:
Hi, the promises do not have to complete simultaneously. They can complete one after the other for all `.all` cares :)

Remember promises are _already started_ operations. Combinators like `.all` that take promises have no control about concurrency - they just combine promises (that's why `.map` and `.filter` which do control concurrency take a function and not a promise).

Issue #1370, opened 2017-04-12
Prioritize bluebird promise resolution over setImmediate callbacks. 


richardpringle:
(This issue tracker is only for bug reports or feature requests, if this is neither, please choose appropriate channel from http://bluebirdjs.com/docs/support.html)

Please answer the questions the best you can:

1) What version of bluebird is the issue happening on?
    3.5.0
2) What platform and version? (For example Node.js 0.12 or Google Chrome 32)
    6.10.0
3) Did this issue happen with earlier version of bluebird?
    yes
(Write description of your issue here, stack traces from errors and code that reproduces the issue are helpful)

Native promise callbacks take precedence over `setImmediate` callbacks. This is not the case with `bluebirdjs`. It would be great if that was also the case with bluebird on Node versions that support native promises. 

Not sure sure how viable this feature request is though...

benjamingr:
You can configure this with `setScheduler` by the way. Opened a PR to discuss.

PR #1374, opened 2017-04-16
Prioritize nextTick over setImmediate

benjamingr:
His aligns the behavior more with native promises.

Discussion https://github.com/petkaantonov/bluebird/issues/1370


petkaantonov:
build fails because of the unused variable

polomsky:
I think the current behaviour is better. Long promise chain does not block IO operations. `setScheduler` can be used to change the behaviour.

PR #1375, opened 2017-04-20
Update promise.all.md

vladimirkosmala:
Added typage of prototype


benjamingr:
Thanks! Landed, @petkaantonov please do a doc rebuild.

PR #1376, opened 2017-04-25
Added asCallback/nodeify to the core

sadasant:
Hi everyone! 🐦 ⚡️ 

My team and I have been needing to use asCallback as a general utility, so I wonder if we could add asCallback/nodeify to the core functions?

Here's a simple PR with a simple test which passes ✅ 💚 

Please let me know if anyone has any form of feedback \o/ and have a wonderful time.

### DIFF
--- a/src/nodeify.js
+++ b/src/nodeify.js
@@ -59,4 +59,18 @@ Promise.prototype.asCallback = Promise.prototype.nodeify = function (nodeback,
     }
     return this;
 };
+
+Promise.asCallback = Promise.nodeify = function(promise) {
+    return function() {
+      var last = arguments.length - 1;
+      var fn;
+      if (last > 0 && typeof arguments[last] === "function") {
+          fn = arguments[last];
+      }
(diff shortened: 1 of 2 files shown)


daedalus28:
In particular, this is useful for methods that don't use bluebird to promisify and only support callback style interfaces so we can just pass promise returning (e.g. async methods) like these lifecycle hooks: http://sailsjs.com/documentation/concepts/models-and-orm/lifecycle-callbacks

With this utility we can do stuff like :

```js
afterCreate: Promise.asCallback(async (val) => await something(val))
```

benjamingr:
I'm currently discussing (last 2 days) on adding this to core: https://github.com/nodejs/CTC/issues/109

Note that in bluebird this would be slow for async/await code since it's native (and not bluebird) promises.

sadasant:
One thing to note is that the current behavior of `nodeify` doesn't pass all the values sent to `resolve` to the callback function, at least that seems to be the case in the test I've been making.

PR #1383, opened 2017-05-10
"I don't care mate" special resolution case

sabarnac:
With reference to [this particular tweet](https://twitter.com/sabarnachk/status/861945503109586944) I made in reply to a Twitter poll, I decided to create a PR after it was suggested that I do so in reply to my tweet. Would love to see this added :smile:.


petkaantonov:
=)

benjamingr:
We actually already have a "I don't care mate" resolution option: cancel :)

PR #1388, opened 2017-05-14
Fix toFastProperties for modern V8

benjamingr:


petkaantonov:
syntax error?

benjamingr:
Oh wait, I ported it from the V8 test but it has to run in older Node, I'll fix this.

benjamingr:
(No syntax error)

PR #1395, opened 2017-05-28
Push to queue in reverse order.

jstans:
Push to queue in reverse order. As .pop() is used for efficiency to pull from the end of the array effectively reversing it again, the end result is a saner execution order. Relates to #708.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -62,7 +62,7 @@ MappingPromiseArray.prototype._promiseFulfilled = function (value, index) {
     } else {
         if (limit >= 1 && this._inFlight >= limit) {
             values[index] = value;
-            this._queue.push(index);
+            this._queue.push(length+limit-index-1);
             return false;
         }
         if (preservedValues !== null) preservedValues[index] = value;
--- a/test/mocha/map.js
+++ b/test/mocha/map.js
@@ -285,10 +285,10 @@ describe("Promise.map-test with concurrency", function () {
             assert.deepEqual(b, [0, 1, 2, 3, 4]);
             lates.forEach(resolve);
         }).delay(100).then(function() {
-            assert.deepEqual(b, [0, 1, 2, 3, 4, 10, 9, 8, 7, 6 ]);
+            assert.deepEqual(b, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9 ]);
             lates.forEach(resolve);
         }).thenReturn(ret1).then(function() {
-            assert.deepEqual(b, [0, 1, 2, 3, 4, 10, 9, 8, 7, 6, 5]);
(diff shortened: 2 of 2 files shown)


spion:
I believe this is deliberately done to avoid relying on execution order when using `map`. The reason for that is because when chaining several maps or map/filter, the execution order will again no longer be guaranteed. Furthermore it may be such that its correct most of the time but not always, creating bugs that only appear one out of 100 times, bugs that only appear in production but not development (due to machine locality) and similar.

Simply forbidding reliance on execution order is the easiest way to prevent all that.

Stanz:
Doesn’t it make more sense to explain this on the documentation rather than make the resulting behaviour confusing in an attempt to... make it less confusing?

I would argue the current implementation is far worse as it goes out of its way to behave differently to how a user would expect a standard map to work. Made even more confusing by working in the correct order with values but reversing the order with promises.

In regards to breaking things. I would argue code should generally not be built to rely on this in any way in the first place as the completion order cannot be guaranteed. But for both prioritisation and real time visualisation it makes sense to at least attempt to run in order... Not do the exact opposite of what you would want in these use cases.

jstans:
Doesn’t it make more sense to explain this on the documentation rather than make the resulting behaviour confusing in an attempt to... make it less confusing?

I would argue the current implementation is far worse as it goes out of its way to behave differently to how a user would expect a standard map to work. Made even more confusing by working in the correct order with values but reversing the order with promises.

In regards to breaking things. I would argue code should generally not be built to rely on this in any way in the first place as the completion order cannot be guaranteed. But for both prioritisation and real time visualisation it makes sense to at least attempt to run in order... Not do the exact opposite of what you would want in these use cases.

spion:
This is mentioned in the documentation

> The order map calls the mapper function on the array elements is not specified, there is no guarantee on the order in which it'll execute the maper on the elements. For order guarantee in sequential execution - see Promise.mapSeries.

tehsenaus:
I'm just coming across this as I've written a tool which makes a lot of API requests with limited concurrency, based on Bluebird's map. I've been puzzled as to why these were requests were happening in reverse order!

I totally your point @spion about providing a fail-fast way of catching potentially obscure bugs though.

To resolve this, may I suggest adding a ```evaluateInOrder``` flag?

That way, as a developer I can assert that yes, I know what I'm doing, and yes, I know order isn't 100% guaranteed, but this is the evaluation behaviour I want.

jstans:
Not really a solution and not optimal for performance but a quick workaround for local code; assuming your array is just promises:

```javascript
var mapInOrder = function(arr,mapper,options) { return Bluebird.map(arr.slice(0,options && options.concurrency || Infinity).concat(arr.slice(options && options.concurrency || Infinity).reverse()), mapper, options) }
```

PR #1411, opened 2017-06-21
Added Examples for await/async

shawngmc:
Added two examples where bluebird works with await/async - using Promisification and Promise.map.

Await/async seems to be really popular at the moment, but Bluebird works well with it - however, without examples, developers may thing bluebird is incompatible or no longer useful.

### DIFF
--- a/docs/docs/features.md
+++ b/docs/docs/features.md
@@ -332,3 +332,42 @@ Your library can then use `var Promise = require("bluebird-extended");` and do w
 
 ##Async/Await
 
+ES2017 brings the async/await concept - which was previously only available as a polyfill - to the Javascript standard. With it, promises are now easier to use than ever: try/catch now works far better and the dreaded 'pyramid code' can be avoided. Bluebird's featureset still helps; two examples are below.
+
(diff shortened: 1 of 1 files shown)


medikoo:
It's also important to note that async functions will always return instance of native `Promise`. So even if internally it's purely bluebird promises that are processed, one should not expect a bluebird promise to be a result of async function.

It might be important to mention that in a doc (I also raised that concern at [ECMAScript discussion list](http://mozilla.6506.n7.nabble.com/Async-functions-not-friendly-to-promise-extensions-td364921.html))

benjamingr:
@medikoo for what it's worth - we already barked down that tree: https://github.com/jhusain/compositional-functions

medikoo:
@benjamingr interesting, but have you tried to voice that concern to TC39? As technically async/await doesn't have to work like that. The only response I got, was something like "because it was simpler to implement in first place"

benjamingr:
@medikoo I've started writing a comment but sent it as an email instead as it contained things better left off GitHub. Cheers.

PR #1419, opened 2017-07-11
Adds additional docs based on 'tap' test suite.

grahamlutz:


benjamingr:
Honestly I don't think that's clearer.

Issue #1426, opened 2017-07-23
Promise.coroutine documentation


liushigit:
I was confused by [the doc](http://bluebirdjs.com/docs/api/promise.coroutine.html) for the `coroutine` method, which says the method

> Returns a function that can use yield to yield promises. 

My understanding of this wording is that, this method "returns a function that may (or may not) yield promises".

But in the provided code example, the returned function is only used as a plain function. There's no "yielded promises" from that function. If this is the example for the "may not" case, shall there be an example for the other case?

benjamingr:
Yeah that documentation is silly. It _takes_ a function that can use yield to yield promises and _returns_ a promise returning function.

A PR would be welcome

liushigit:
Here's the `coroutine` method's [source code](https://github.com/petkaantonov/bluebird/blob/master/src/generators.js#L191). 

The first sentence may be changed to "Takes a [`GeneratorFunction`](https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Iterators_and_Generators) producing a generator that yields promise(s), and returns a `Promise` returning function."

The returned `Promise` is defined at [this line(https://github.com/petkaantonov/bluebird/blob/master/src/generators.js#L43), but I don't understand its use.

benjamingr:
@liushigit I'll clarify, I agree that the docs are confusing (and wrong), and if you want to submit a pull request I promise to review it and merge it.

liushigit:
@benjamingr A PR is added, but I have difficulty explaining the `Promise` returned by the function returned.

benjamingr:
@liushigit thanks! I've added some suggestions, LMK what you think

perrin4869:
I'm not sure if this is the right place to ask this question, but what's the difference performance-wise or other, between using coroutines and async/await in node >= 8? Maybe a note should be added to the documentation about it?

benjamingr:
@perrin4869 it's likely async/await is still slower than `Promise.coroutine` at the moment. You can run the bluebird doxbee benchmarks yourself.

PR #1428, opened 2017-07-24
Update promise.coroutine.md

liushigit:
This is for this [issue](https://github.com/petkaantonov/bluebird/issues/1426)

TODO: The `Promise` by the function returned by the `coroutine` method needs further explanation. A thorough understanding of [the code](https://github.com/petkaantonov/bluebird/blob/4f9093448f55ea76d5a8e090e42fe24b8e0da82c/src/generators.js#L191) is needed.

### DIFF
--- a/docs/docs/api/promise.coroutine.md
+++ b/docs/docs/api/promise.coroutine.md
@@ -13,7 +13,9 @@ title: Promise.coroutine
 Promise.coroutine(GeneratorFunction(...arguments) generatorFunction, Object options) -> function
 ```
 
-Returns a function that can use `yield` to yield promises. Control is returned back to the generator when the yielded promise settles. This can lead to less verbose code when doing lots of sequential async calls with minimal processing in between. Requires node.js 0.12+, io.js 1.0+ or Google Chrome 40+.
+In a typical use case, the parameter `generatorFunction` should be a `GeneratorFunction` producing a generator that yields one or more `Promise`s. (Also see the Tips section below.)
+
+This method returns a function that returns a `Promise`(TODO: Elaborate this). When the returned function (e.g. the `ping` function in the example below) is called, it will start the generator (by calling its `next()`) and return immediately. Control is returned back to the generator when each yielded promise settles. This can lead to less verbose code when doing lots of sequential async calls with minimal processing in between. Requires node.js 0.12+, io.js 1.0+ or Google Chrome 40+.
 
 ```js
 var Promise = require("bluebird");
@@ -51,8 +53,6 @@ Running the example:
     Pong! 7
     Ping? 8
     ...
-
-When called, the coroutine function will start an instance of the generator and returns a promise for its final value.
 
 Doing `Promise.coroutine` is almost like using the C# `async` keyword to mark the function, with `yield` working as the `await` keyword. Promises are somewhat like `Task`s.
 


docs/docs/api/promise.coroutine.md
```diff
@@ -13,7 +13,9 @@ title: Promise.coroutine
 Promise.coroutine(GeneratorFunction(...arguments) generatorFunction, Object options) -> function
 ```
 
-Returns a function that can use `yield` to yield promises. Control is returned back to the generator when the yielded promise settles. This can lead to less verbose code when doing lots of sequential async calls with minimal processing in between. Requires node.js 0.12+, io.js 1.0+ or Google Chrome 40+.
+In a typical use case, the parameter `generatorFunction` should be a `GeneratorFunction` producing a generator that yields one or more `Promise`s. (Also see the Tips section below.)
```
benjamingr: Nit, the generator is the function itself - it returns an iterator. So it's a `GeneratorFunction` but it's not producing a generator function.

Maybe: "In a typical use case, the parameter `generatorFunction` should be a `GeneratorFunction` that can yield zero or more `Promise`s. (Also see the Tips section below.)"?

docs/docs/api/promise.coroutine.md
```diff
@@ -13,7 +13,9 @@ title: Promise.coroutine
...
+
+This method returns a function that returns a `Promise`(TODO: Elaborate this). When the returned function (e.g. the `ping` function in the example below) is called, it will start the generator (by calling its `next()`) and return immediately. Control is returned back to the generator when each yielded promise settles. This can lead to less verbose code when doing lots of sequential async calls with minimal processing in between. Requires node.js 0.12+, io.js 1.0+ or Google Chrome 40+.
```
benjamingr: About the todo: 

"That returns a Promise that resolves to the value `return`ed from the generator function (if that value is a promise, it will be unwrapped)" ?

That's not entirely accurate (since a finally block can still intercept it - but I think it's accurate _enough_ for virtually 100% of user use cases.

docs/docs/api/promise.coroutine.md
```diff
@@ -52,8 +54,6 @@ Running the example:
     Ping? 8
     ...
 
-When called, the coroutine function will start an instance of the generator and returns a promise for its final value.
```
benjamingr: Why is this removed?

benjamingr:
Hey, thanks a lot for this! Left some feedback on the PR.

docs/docs/api/promise.coroutine.md
liushigit: It is replaced by "When the returned function (e.g. the `ping` function in the example below) is called, it will start the generator (by calling its `next()`) and return immediately."

docs/docs/api/promise.coroutine.md
liushigit: Ah. I thought it's the generator that "yields", actually it's the `GeneratorFunction`, which can be seen as the "factory".

docs/docs/api/promise.coroutine.md
liushigit: - What will the promise be like when multiple values or promises are yielded?
- We can use `await` now, so that kind of use case will be very rare. Also, the example didn't make use of the returned promise.

benjamingr: > What will the promise be like when multiple values or promises are yielded?

`yield`ed values don't affect the return value, the promise will always be the `return` value.

> We can use await now, so that kind of use case will be very rare.

I agree, `Promise.coroutine` can be used as a performance feature for code after you prove it's needed. Although [@spion disagrees](https://spion.github.io/posts/es7-async-await-step-in-the-wrong-direction.html).

> Also, the example didn't make use of the returned promise.

An example would definitely be nice here.

docs/docs/api/promise.coroutine.md
benjamingr: It `return`s a function that when called runs the generator function. The factory here is `Promise.coroutine` and the generator is run exactly once per invocation (and pumped in the code).

liushigit: Just to be clear, [a generator is not a `GeneratorFunction`, and is an "iterator"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) :

>  The Generator object is returned by a generator function and it conforms to both the iterable protocol and the iterator protocol.

benjamingr:
Hey, are you going to address the feedback and finish this @luishigit

liushigit:
@benjamingr Hi sorry for the delay. Now I think there should also be a summary for the purpose of this function. It returns a ["coroutine"](https://en.wikipedia.org/wiki/Coroutine), which is something can be executed "concurrently" and often, taken care of by an "event loop". The `yield` in the function as the argument has a similar effect as the [`await` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) . 

Some thing like that would be helpful.

docs/docs/api/promise.coroutine.md
Retsam: > Requires node.js 0.12+, io.js 1.0+ or Google Chrome 40+.

Could this be expanded, perhaps?  It'd certainly be good to know what IE support looks like, and whether this can be supported by a polyfill, or if there's a more fundamental limitation.  (In the comments on the page, I see someone who says that a polyfill enables it for IE10, but not IE9, but I see no confirmation of that)

I can dig into this, but I'm guessing someone more knowledgeable might know off-hand.

Issue #1404, opened 2017-06-07
returning Promise.reject in async function results in Unhandled rejection error message


comtaler:
1) What version of bluebird is the issue happening on?
3.5.0

2) What platform and version? (For example Node.js 0.12 or Google Chrome 32)
Node.js 8.0.0 on Ubuntu

3) Did this issue happen with earlier version of bluebird?
Yes

When returning Promise.reject in async function, an error is printed on the console "Unhandled rejection" even the promise is caught. It doesn't print out the error if native Promise is used.
```
const Promise = require("bluebird");

async function WaitAsync(){
    return Promise.reject(new Error("reject"));
}

Promise.resolve().then(() => {
    return WaitAsync();
}).catch(err => {
    console.log("caught: ", err);
});
```

sorccu:
I believe I'm encountering another instance of this issue. In fact I was going to post a new issue before I saw this one. Here's my description:

This issue is a bit difficult to grasp, so I'll do my best to explain with examples. It seems that when using `Promise.using()` with an async function, awaiting on already rejected promises always outputs an `Unhandled rejection` when using Bluebird's `Promise.reject()`, but not when using native `Promise.reject()` or `throw` in an async function.

These code snippets are runnable directly on node 8 (tested on v8.1.0) with Bluebird v3.5.0 (and earlier).

First, let's start from a problematic example.


```js
// Failing example 1

const Promise = require('bluebird')

Promise.using(Promise.resolve(), async () => {
  await Promise.reject(new Error('foo'))
})
.catch(() => console.log('OK'))
```

This snippet, when run, produces the following output:

```
Unhandled rejection Error: foo
    at Promise.using (<redacted>/a.js:4:24)
    at tryCatcher (<redacted>/node_modules/bluebird/js/release/util.js:16:23)
    at <redacted>/node_modules/bluebird/js/release/using.js:185:26
    at tryCatcher (<redacted>/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (<redacted>/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (<redacted>/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (<redacted>/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (<redacted>/node_modules/bluebird/js/release/promise.js:693:18)
[...]

PR #1430, opened 2017-08-02
avoid unexpected unhandledRejection

suguru03:
I fixed this issue. https://github.com/petkaantonov/bluebird/issues/1404
Could you review it?

I wanted to add these cases, but there is no `async/await` test case. Therefore I didn't. (If you want, I will add them..)

### example 0

```js
async function WaitAsync(){
    return Promise.reject(new Error("reject"));
}

Promise.resolve().then(() => {
    return WaitAsync();
}).catch(err => {
    console.log("caught: ", err.message);
});
```
```
caught:  reject
```

### example 1
```js
Promise.using(Promise.resolve(), async () => {
  await Promise.reject(new Error('foo'))
})
.catch(() => console.log('OK'))
```

```
OK
```

### example 2
```js
const reject = () => Promise.reject(new Error('foo'))

Promise.using(Promise.resolve(), async () => {
  await reject()
})
.catch(() => console.log('OK'))
```

```
OK
```

### DIFF
--- a/src/debuggability.js
+++ b/src/debuggability.js
@@ -37,9 +37,12 @@ Promise.prototype.suppressUnhandledRejections = function() {
 };
 
 Promise.prototype._ensurePossibleRejectionHandled = function () {
-    if ((this._bitField & IS_REJECTION_IGNORED) !== 0) return;
-    this._setRejectionIsUnhandled();
-    async.invokeLater(this._notifyUnhandledRejection, this, undefined);
+    var promise = this;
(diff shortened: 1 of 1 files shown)


STRML:
@petkaantonov Please take a look - this is causing spurious logging in production applications.

petkaantonov:
Thanks, same fix was already implemented. However it causes more issues because setTimeout is hijacked by some testing frameworks.

PR #1451, opened 2017-09-11
Breaking Promiss chains - Proposal - SkipTo/Resume/ResumeOrContinue  

rubens21:
Hey guys,
I had found some discussions on internet about a nice way to break a promise chain. Then I decided to share a solution that I am using for a while.
I hope it may be useful for someone else.
This Pr creates a Promise.skipTo static method, that allows us to skip as many "then" necessary to reach a Resume/ResumeOrContinue  point, and then continuing from there.


benjamingr:
I really wish you would have talked to us before you started all this work.

For what it's worth the "common" name for `skipTo` is `goto` and for `resume` it's "label".

rubens21:
Thanks @benjamingr,
I may change it.
I will talk to you guys next time ;-)

rubens21:
By the way, do you have a suggestion for the method "resumeOrContinue" 
Differently of "resume", the "resumeOrContinue" is also executed as a "then" block. It means that even if there is no "goto" point called this block is going to be executed.

rubens21:
@benjamingr I tried to change the name to `goto`, however it was not working well because that is a key word.
image

I, personally, prefer the word `skip` because we are skipping some points/links in a chain. While goto make think about a "position". Whatever, actually does not matter too much :-).

I do not know if the Bluebird team are interested by this feature. If not, not feelings, just reject the PR.

Issue #984, opened 2016-01-28
Feature Request: Branching construct


erg0dic:
Frequently I find myself wanting to add some control flow to my promise chains, e.g:

```javascript
let startSomething = () => {
  // This promise resolves with a value & some set of conditions about what happens next
};

let aChain = doSomething().then(finishSomething);
let bChain = doSomethingElse.then(finishSomethingElse);
let cChain = doAnotherThing.then(finishAnotherThing);

// n = 2
// where predicateFn = (Any -> Boolean)
startSomething().then(Promise.decide(predicateFn, aChain, bChain))
// if predicateFn is true, we execute aChain with the value resolved from startSomething, 
// else we do the same with bChain

// For all n
// where predicateFnStar = (Any -> Listof[Boolean])
// Invariant: The list of booleans returned from predicateFnStar must not contain
// more than one true value
doSomething.then(Promise.select(predicateFnStar, aChain, bChain, cChain, ... , nChain))
// The structure of the array returned by predicateFnStar determines which of aChain,
// bChain, &c is executed with the value resolved from startSomething
```

Is this something you could see in bluebird?

Best,
Rob

benjamingr:
Promises are already started values. In your case, both aChain, bChain and cChain area already executing. 

Control flow can be accomplished in a variety of ways, namely an `if/else` works :)

Celadora:
Is this a feature you would consider were someone were to make a pull request? We have a solution that we'd like to offer.

benjamingr:
Feel free to offer the solution, no guarantees

Celadora:
```
  var Promise = require('bluebird');
  Promise.prototype.branch = function(...fns) {
   return this.all().spread((index, value)=>{
    if(index===null ||
       index === undefined ||
       typeof fns[index] !== 'function') return Promise.reject(value);
    return fns[index](value);
   })
  }
  
  new Promise.resolve().then(nothing=>{
    return [2, 'somevalue']
  })
  .branch(a, b, c, d)
  .done(value=>{
    console.log(value);
  })
  function a(value) {
    console.log('a', value);
    return 0;
  }
  function b(value) {
    console.log('b', value);
    return 1;
  }
  function c(value) {
    console.log('c', value);
    return 2;
  }
  function d(value) {
    console.log('d', value);
    return 3;
  }
```

It's used much like spread, in that you return an array from a previous link in the chain. The array is simple: [index, value]. The index corresponds to the branch, and the value is arbitrary.

[...]

PR #1454, opened 2017-09-14
Add Conditional/Branching construct

justsml:
This is a proposed change related to #984 
And my comments/examples here: https://github.com/petkaantonov/bluebird/issues/984#issuecomment-328633367

## API 
```js
Promise.thenIf(condition(value), ifTrue(value), ifFalse(value))
```

## Usage
```js
// Use like so:
const checkEmail = email => Promise.resolve(email)
  .thenIf(
    e => e.length > 5,              // Conditional
    e => console.log('Valid: ', e), // ifTrue
    e => console.error(e))          // ifFalse
// Or like so:
const checkEmail = email => Promise.resolve(email)
  .thenIf(e => e.length > 5)
// Or:
const checkEmail = email => Promise.resolve(email)
  .then(Promise.thenIf(e => e.length > 5))
```

Lots of cool stuff is possible, the default values let you call `.thenIf` with no args - if you simply want to exclude falsey values down the chain.


justsml:
Hmmm... Is this a travis fail?
image

Tests are passing in all my local tests for node v6-v8.
image

spion:
As in the closed issue, usually we recommend writing a `then` combinator for these cases:

```es6
Promise.resolve(email)
.then(iff(e => e.length > 5,         // Conditional
    e => console.log('Valid: ', e),  // ifTrue
    e => console.error(e)))          // ifFalse
```

Main benefit being that this will work with any promise library.

justsml:
I support that pattern (exposed as `Promise.thenIf`) - I've been using it in almost all my projects for a few months now.
Considering the array methods, I feel branching is a natural addition.

Mainly I hate having to patch bluebird with my util/helpers. (tired of disclaiming - "don't ever normally hack onto native prototypes...") You might ask, Well then, "why?" Is this scope creep? 

I'd say bluebird is already a batteries-included library. Adding something that seems more relevant than `.delay()` is at least worthy of some more consideration.

justsml:
@spion that is really close, but if I wanted to write my function chains that way I'd use https://github.com/sindresorhus/ (still awesome) 1000 mini promise libs. 😉

justsml:
Here's the ES6 I'm working off, hopefully easier to read/review:

```js
module.exports = function _conditionals(Promise) {
  Promise.prototype.thenIf  = thenIf
  Promise.thenIf            = _thenIf

  function thenIf(cond = (x) => x, ifTrue = (x) => x, ifFalse = () => null) {
    if (this instanceof Promise) {
      return this.then(value => _thenIf(cond, ifTrue, ifFalse)(value))
    } else {
      return _thenIf(cond, ifTrue, ifFalse)
    }
  }

  function _thenIf(cond = (x) => x, ifTrue = (x) => x, ifFalse = () => null) {
    return value => Promise.resolve(cond(value))
      .then(ans => ans ? ifTrue(value) : ifFalse(value))
  }
}

```

http://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=LYewJgrgNgpgdDAHgBxAJwC4GcAEBeHAMwgDsBjDASxBJwH0yaxKqaBDKLACgAU0RglLDACUOAN4AoHDj4Ch8ZPwwgMAT2TwMACxgkAkoRkEdew9Nn9BwuKYNGZjp8fp3zF4uVa03hLoxIwfBwuRDE8AD4cRAAaHEpCABU0CBhg0PCo2PjCADEOYXTMnBJoKDEpJwSQnSF4kiwMNnIYECM5a1EJCyc0GAwINB9tIVtdEi4ANw5U_Ci6X38mOITk1JW8gtEpmdERHpwAXxwYTjTK5z6BoddxwyXAjbWYDfyz_adDiy-PUgpqWgLO5-AJBAgZObRJ4pNLgsKQ7IJN6FcHFUpQcrdXr9Qa0aZQWaRSzyGx9LAgKCTGAPMA7Al7A4yMZ6LjNXBEtk4AD8OWedNSYgAXDlkdT8QKPkdJF8gA&debug=false&circleciRepo=&evaluate=false&lineWrap=false&presets=es2015%2Ces2017%2Creact%2Cstage-2%2Cstage-3&prettier=true&targets=&version=6.26.0

spion:
Bluebird's external API is currently closed-by-default, unless there is a significant new construct that provides new functionality (e.g. using, streams, etc). API that can be achieved by adding a combinator is generally not added anymore since we find that the large API surface area burden on users outweighs the gain

This sucks, but unfortunately JS's OO flavour doesn't make it easy to separate extensions from a library, so either everyone is burdened with a large surface area, or noone gets all the features they want. Maybe [the bind operator](https://github.com/tc39/proposal-bind-operator) would make things better, if only TC39 got their act together in pushing it through.

@benjamingr @petkaantonov correct me if I'm wrong.

justsml:
I keep finding examples where `.tapIf()`/`.thenIf()` would clean up code very nicely IMHO (by eliminating extra wrapping clousures/arrow funcs.)

```js
const ensureDirectory = (file) => Promise
  .resolve(path.join(__dirname, '..', file.category))
  .tapIf(fs.existsAsync, fs.mkdirAsync) // <-- oh daayyyum!!!
  .then(() => file) 
// Will create folder path (if needed), then return the input `file`
```

justsml:
@spion 
Regardless of the feature relevance, I'd really appreciate some help figuring out the error in tests, `late_buffer_safety.js` `throw`'s integers, I thought that wasn't supported? 😕 

And how on earth does my code affect this unrelated file? 😖 

Thanks!

jonathanherring:
Seems like a good way to easily add if/else conditions. 👍

Issue #990, opened 2016-02-03
Warning: a promise was rejected with a non-error: [object Error]


ColCh:
A warning is fired when I reject a Promise with an `new Error('...')`, but __Promise__ and __Error__ are from different `Window` 's (e.g. __window.opener__ and *window child* - from __window.open__ or __iframe.contentWindow__).

Of cource, it means, that `rejectReason instanceof Error` returns `false`, because __Error__ is instance of __iframe.contentWindow.Error__ (not of __window.Error__)

ColCh:
__Upd__: To hot-fixing this issue.

You can create another instance for __Error__ and proxy all props from previous to new, and *reject* a promise with created *Error*.
```
            if (err) {
                const proxiedError = new Error();
                proxiedError.message = err.message;
                proxiedError.stack = err.stack;
                reject(proxiedError);
            } else {
                resolve(res);
            }
```

However, *bluebird* should use __[[Class]]__ prop to determine errors (`Object.prototype.toString`), isn't it? Or I'm wrong?

benjamingr:
What about error subclasses? 
Honestly I'm fine with checking if it's possible to attach a stack to it.

ColCh:
For subclasses - same. Checked with `EvalError` and `ReferenceError`.

benjamingr:
Let's be more concrete. Can you show me a piece of code that takes something and determines if it's an error?

ColCh:
Sure. Consider following snippet:
```js
export async function evalInWindowAsync (win, func/*nodeClb, jQuery*/, additionalArgs = []) {
    return new Promise((resolve, reject) => {

        let args = [];
        const argsId = _.uniqueId('evalInWindow_arg_holder_');

        const callback = function(err, res) {

            delete win[argsId];

            if (err) {
                const proxiedError = new EvalError();
                proxiedError.message = err.message;
                proxiedError.stack = err.stack;
                reject(proxiedError);
            } else {
                resolve(res);
            }
        };

        args.push(callback);
        args = args.concat(additionalArgs);
        win[argsId] = args;

        const evalArgs = [`window["${argsId}"][0]`, `window.jQuery`];

        for (var i = 1; i < args.length; i++) {
            evalArgs.push(`window["${argsId}"]["${i}"]`);
        }

        const functionString = _.isString(func) ? functionString : func.toString();
[...]

PR #1463, opened 2017-10-03
Better error check

gabegorelick:
See https://github.com/petkaantonov/bluebird/issues/990#issuecomment-333598787


petkaantonov:
Thanks

PR #1464, opened 2017-10-05
Conform to spec when generating PromiseRejectionEvent

### LINKED ISSUE #1447: PromiseRejectionEvent differs from Living Standard spec
3.5.0

Google Chrome 56.0

Yes.

Bluebird seems to be creating a PromiseRejectionEvent with a 'detail' property, which contains 'promise' and 'reason' properties. TypeScript complains that PromiseRejectionEvent has no 'detail' property, which prompted me to look at the specification for this event. It appears that 'promise' and 'reason' are intended to be properties of the event object itself, e.g. event.promise and event.reason.

Perhaps I'm missing something?

Spec info found here -
https://developer.mozilla.org/en-US/docs/Web/API/PromiseRejectionEvent/PromiseRejectionEvent
https://html.spec.whatwg.org/multipage/webappapis.html#promiserejectionevent

nornagon:
Fixes #1447

### DIFF
--- a/src/debuggability.js
+++ b/src/debuggability.js
@@ -4,6 +4,7 @@ var getDomain = Promise._getDomain;
 var async = Promise._async;
 var Warning = require("./errors").Warning;
 var util = require("./util");
+var es5 = require("./es5");
 var ASSERT = require("./assert");
 var canAttachTrace = util.canAttachTrace;
 var unhandledRejectionHandled;
@@ -154,16 +155,13 @@ var fireDomEvent = (function() {
             var event = new CustomEvent("CustomEvent");
             util.global.dispatchEvent(event);
             return function(name, event) {
-                var promise = event.promise;
-                var reason = event.reason;
                 var eventData = {
                     detail: event,
                     cancelable: true
                 };
-                Object.defineProperties(eventData, {
-                    promise: { get: function() { return promise; } },
-                    reason: { get: function() { return reason; } }
-                });
+                es5.defineProperty(
+                    eventData, "promise", {value: event.promise});
+                es5.defineProperty(eventData, "reason", {value: event.reason});
                 var domEvent = new CustomEvent(name.toLowerCase(), eventData);
                 return !util.global.dispatchEvent(domEvent);
             };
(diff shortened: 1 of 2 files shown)


src/debuggability.js
```diff
@@ -156,7 +156,9 @@ var fireDomEvent = (function() {
             return function(name, event) {
                 var domEvent = new CustomEvent(name.toLowerCase(), {
                     detail: event,
-                    cancelable: true
+                    cancelable: true,
+                    reason: event.reason,
+                    promise: event.promise
```
nornagon: Would it make sense to write this as `get promise() { return event.promise; }`? I'm unfamiliar with the compilation pipeline for bluebird, and whether that syntax is supported on all the browsers Bluebird targets.

If we wanted to be even more protective, we could do something like:
```js
var reason = event.reason;
var promise = event.promise;
var domEvent = new CustomEvent(name.toLowerCase(), {
  detail: event,
  cancelable: true,
  get reason() { return reason; },
  get promise() { return promise; }
})
```

or the equivalent with `Object.defineProperty`

Let me know if that's a desirable change & I can update the PR

nornagon:
I just realised this doesn't handle the case where CustomEvent isn't defined, will update.

src/debuggability.js
benjamingr: Bluebird actually targets ES3 browsers too (supports IE7 for example) - on the other hand this branch of code isn't reached on these browsers anyway.

nornagon:
Updated to use getters. Let me know if I should squash these commits.

src/debuggability.js
```diff
@@ -154,12 +154,17 @@ var fireDomEvent = (function() {
...
+                var promise = event.promise;
+                var reason = event.reason;
+                var eventData = {
                     detail: event,
-                    cancelable: true,
-                    reason: event.reason,
-                    promise: event.promise
+                    cancelable: true
+                };
+                Object.defineProperties(eventData, {
```
benjamingr: If you're using this syntax anyway - can you please use the helpers from util that work in ES3 environments for consistency?

nornagon: Done (I think!)

petkaantonov:
so is it supposed to be data property or getters? it's currently data property

petkaantonov:
checked the spec and its readonly data.. merging

PR #1467, opened 2017-10-12
Create await-all.js

seitbekir:
This function is something like `Promise.all()`, but wait for all promises, and if one or more of them are failed -- reject array of error results. If all are fine -- same, but resolve (same as `.all()`).

PS: I'll add tests for the function later. Just keep this PR for this day here.

### DIFF
--- a/src/await-all.js
+++ b/src/await-all.js
@@ -1,0 +1,30 @@
+"use strict";
+module.exports = function(
+    Promise
+) {
+    function awaitAll(promises) {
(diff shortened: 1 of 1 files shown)


benjamingr:
You can just use `.settle` or better - `.all` with `.reflect`:

```js
  let results = Promise.all(values.map(x => fn(x).reflect()))
  // results is a list of PromiseInspection here and you can filter the resolved/rejects 
```

Issue #1161, opened 2016-07-12
Using bluebird in IE/Edge does not always resolve promises when window is not focused


mindlink:
1) What version of bluebird is the issue happening on?
3.4.1

2) What platform and version? (For example Node.js 0.12 or Google Chrome 32)
Internet Explorer 11, Edge

3) Did this issue happen with earlier version of bluebird?
Unknown

Have some operation that triggers regardless of user interaction e.g. long-polling XHR, websocket, setTimeout, that signals a promise chain and nothing else. Then switch from IE/Edge to something else.

You would expect that the promise chain is executed regardless, however the chain stops executing.

Further investigation shows that the use of MutationObserver as the scheduler, means that a DOM change (or console log if the debugger is open) is required to trigger the promise chain. A detached DOM operation does not trigger the chain (like in the MutationObserver scheduler).

We have worked around this issue by setting the scheduler to use setTimeout (which isn't ideal).

benjamingr:
We actually use `NativePromise.resolve` if it is available so edge shouldn't be an issue.

As for IE11, I haven't experienced this - is the computer under low power mode or something similar? Note that in low power settings your `setTimeout` workaround might not work. 

We can prefer setImmediate over mutationobserver in IE. A small reproducing example would do wonders.

lowkay:
Howdy,

Further investigation into this revealed that the issue is only in a callback from an XHR. Resolving a promise in such a callback without active DOM operations will not trigger the promise chain to execute.

We thus altered our approach to wrap the promise resolution in a setTimeout, then magically it works.

I have a test case, which I'll try and publish on github next week. It seems to affect pretty much every promise library not just bluebird - due to the fact MutationObserver has the closest execution scheduling to native promises.

benjamingr:
Oh, so it's an XHR specific issue in IE where it doesn't call the resolution code unless it's wrapped in a timeout? That's interesting.

I don't think it's in scope for bluebird to solve that case though as it's xhr specific.

thomas-darling:
@benjamingr can you please elaborate on why this not a bluebird issue?
This is currently prevents our app from loading in IE, and I don't see who else would be in a position to fix this - as I understand it, this is a result of the way bluebird uses MutationObserver for scheduling.

benjamingr:
> Further investigation into this revealed that the issue is only in a callback from an XHR. Resolving a promise in such a callback without active DOM operations will not trigger the promise chain to execute.

[...]

PR #1476, opened 2017-11-07
disable mutation-observer scheduler in IE

davedoran:


benjamingr:
@davedoran hey

davedoran:
Yeah - sorry; didn't mean to submit this PR back to the main fork.

benjamingr:
@davedoran ok - if I may ask - why was it required on your end?

(You can `setScheduler` by the way and avoid changing the source code)

davedoran:
I develop a large application that is migrating from jQuery to BlueBird.  We're noticing that - on IE11 - the MutationObserver occasionally fails to trigger when it should. The result is (rarely) a serious performance impact on IE. We have seen delays in the range of minutes where the application is blocked waiting for the callback. 

Unfortunately, I don't have a reliable testcase outside of my application (hence - not submitting this patch) but falling back to `setTimeout` on IE sidesteps the issue. 

Thanks for the hint re: ```setScheduler```.  That will be very useful.

benjamingr:
Thanks for the update, we'll look into it.

davedoran:
Cool - thanks.  I suspect we're hitting the same issue discussed here: 

https://github.com/petkaantonov/bluebird/issues/1161
https://github.com/lowkay/MutationObserverSchedulerBug

PR #1477, opened 2017-11-09
Make it more clear that .all is parallel

freeall:
This goes hand in hand with https://github.com/petkaantonov/bluebird/blob/master/docs/docs/api/promise.each.md where it's stated that iteration is serial.

### DIFF
--- a/docs/docs/api/promise.all.md
+++ b/docs/docs/api/promise.all.md
@@ -15,7 +15,7 @@ Promise.all(Iterable<any>|Promise<Iterable<any>> input) -> Promise<Array<any>>
 
 This method is useful for when you want to wait for more than one promise to complete.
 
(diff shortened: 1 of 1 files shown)


benjamingr:
There is no iteration and it doesn't happen in parallel (it happens concurrently).

A promise is _the result_ of an action _over time_, this means that once you pass an array of promises to `Promise.all` _all the actions have already started_ before the call to `.all`. 

It would be _impossible_ to even build a `.all` where the promises aren't concurrent since by the time you're passing promises it's "too late", that's why methods that work with concurrency settings take functions and not promises.

To be clear:

```js
var p1 = somePromiseReturningFn();
var p2 = somePromiseReturningFn();
// at this point, all actions are already running, 
Promise.all([p1, p2]); // this line has no impact on how concurrent the action is or is not
```

`Promise.all` only takes an array of promises and returns a promise for an array when all of them resolve - it doesn't start or run anything.

Thanks a lot for the contribution though, we appreciate it.

PR #1481, opened 2017-11-24
Fix type annotation for Promise.some in docs

shofel:


shofel:
I see, I've misunderstood. I was just confused with `<any>` part of the type spec. `Any` means here "a value or a Promise that resolves to a value"

Should this be 

type iterableOrPromiseT<A> = Iterable<A | Promise<A>>
type promiseSomeT<A> = (iterableOrPromiseT<A>, count: number) => Promise<*>

Not so easy.. Maybe some day will write typings for Flow.

PR #1489, opened 2017-12-20
Fix Promise.map unhandled rejections (#1487)

### LINKED ISSUE #1487: Unhandled rejection from `.map()` and `.filter()` in v3.5.1

3.5.1

Node 6.12.2, Mac OS 10.12.6

No. Issue not present in bluebird 3.5.0.

---

Chaining `.map()` or `.filter()` on to a rejected promise leads to an unhandled rejection, even when `.catch()` handler is attached further down the chain.

```js
Promise.reject( new Error('foo') )
    .map( () => {} )
    .catch( () => {} );
```

With bluebird v3.5.1, the above produces `Unhandled rejection Error: foo`. In v3.5.0 no unhandled rejection is created.

I assume is caused by https://github.com/petkaantonov/bluebird/commit/48c8591c0ad3fc6a8f6dc38956fe6a08dd0b617f but I can't understand why!

Don't know if this is related to https://github.com/petkaantonov/bluebird/issues/1468.

overlookmotel:
This PR fixes #1487.

Previously, unhandled rejection errors can occur with `Promise.map()`, `Promise.filter()`, `.map()` and `.filter()` when in fact the rejections are handled.

Prior to this PR, all of the following create an unhandled rejection, after this PR they do not.

```js
var p = Promise.reject( new Error('foo') );
Promise.map( p, () => {} ).catch( () => {} );

var arr = [ Promise.reject( new Error('foo') ) ];
Promise.map( arr, () => {} ).catch( () => {} );

var p = Promise.reject( new Error('foo') );
p.map( () => {} ).catch( () => {} );

var arr = [ Promise.reject( new Error('foo') ) ];
var p = Promise.resolve( arr );
p.map( () => {} ).catch( () => {} );


var p = Promise.reject( new Error('foo') );
Promise.filter( p, () => {} ).catch( () => {} );

var arr = [ Promise.reject( new Error('foo') ) ];
Promise.filter( arr, () => {} ).catch( () => {} );

var p = Promise.reject( new Error('foo') );
p.filter( () => {} ).catch( () => {} );

var arr = [ Promise.reject( new Error('foo') ) ];
var p = Promise.resolve( arr );
p.filter( () => {} ).catch( () => {} );
```

I have not written tests as I'm not sure how to test for unhandled rejections.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -23,13 +23,14 @@ function MappingPromiseArray(promises, fn, limit, _filter) {
     this._limit = limit;
     this._inFlight = 0;
     this._queue = [];
-    async.invoke(this._asyncInit, this, undefined);
+    this._asyncQueue = [];
+    this._init$(undefined, RESOLVE_ARRAY);
+    if (!this._isResolved() && this._asyncQueue.length > 0) {
+        async.invoke(this._drainAsyncQueue, this, this._asyncQueue);
+    }
+    this._asyncQueue = undefined;
 }
 util.inherits(MappingPromiseArray, PromiseArray);
-
-MappingPromiseArray.prototype._asyncInit = function() {
-    this._init$(undefined, RESOLVE_ARRAY);
-};
 
 // The following hack is required because the super constructor
 // might call promiseFulfilled before this.callback = fn is set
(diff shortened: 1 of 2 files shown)


benjamingr:
Can you add tests?

What about two rejections in a single array?

overlookmotel:
Would be happy to write tests, but can you advise how to write a test for an unhandled rejection? By it's nature it's hard to test for.

2 rejections in a single array should also be handled as iteration over the array now happens synchronously.

overlookmotel:
@benjamingr Tests added. I've used `._isRejectionUnhandled()` to check that the rejections are handled synchronously. Seems a bit wonky to be using private undocumented APIs in tests, but I can't see any other reasonable way to do it.

overlookmotel:
NB The test also covers case of 2 rejections in an array.

Admasnd:
Has this been merged to master yet?

PR #1493, opened 2018-01-06
Remove dead code in Queue.prototype.push

overlookmotel:
This PR removes one line of unnecessary code in `queue.js`.

`this._checkCapacity(length)` will never do anything as the eventuality where queue is over capacity has already been checked for by `this._willBeOverCapacity(length)` a few lines earlier.

### DIFF
--- a/src/queue.js
+++ b/src/queue.js
@@ -38,7 +38,6 @@ Queue.prototype.push = function (fn, receiver, arg) {
         return;
     }
     var j = this._front + length - 3;
-    this._checkCapacity(length);
     var wrapMask = this._capacity - 1;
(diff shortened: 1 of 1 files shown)


rembogaming633-spec:
What sorcery is this?!

so just dont wwrite cryptic ancient js that nobody knows because turbo doesnt know eitherhttps://github.com/petkaantonov/bluebird.wiki.git

rembogaming633-spec:
What sorcery is this?!

so just dont wwrite cryptic ancient js that nobody knows because turbo doesnt know eitherhttps://github.com/petkaantonov/bluebird.wiki.git

Uploading f40e4fa8bef0791fcdb474d02184f42b.mp4…

PR #1530, opened 2018-05-31
Update README.md

benjamingr:
@petkaantonov @bmeurer @MayaLekova PTAL


benjamingr:
LGTM from Petka who doesn't have access to his phone right now for 2fa, if anyone wants to propose improvements we can :)

PR #1539, opened 2018-07-16
Update benchmarks

kjnsn:


benchmark/README.md
```diff
@@ -1,72 +1,77 @@
...
-    Intel(R) Core(TM) i5-2500K CPU @ 3.30GHz × 4
+    Darwin 17.7.0 x64
+    Node.JS 8.11.1
+    V8 6.2.414.50
+    Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz × 4
 
 bench parallel (`--p 25`)
 
-results for 10000 parallel executions, 1 ms per I/O op
+results for 10000 parallel executions, 1 ms per I/O op `ls ./madeup-parallel/*.js | sed -e 's|\.js||' | xargs node ./performance.js --p 25 --t 1 --n 10000`
```
benjamingr: I'm not sure I understand this command - `ls ./madeup-parallel/*.js` ?

benjamingr:
Can you run this with Node.js 10?

benchmark/README.md
kjnsn: The assumption is that the command would be run from the `benchmark` directory

PR #1542, opened 2018-07-24
Guard against falsey process.versions to support React Native

L8D:
I ran into this issue when working on a React Native project. When I imported bluebird, it would throw `Cannot read property 'node' of undefined` and I had to work around the problem by using `Object.defineProperty(process, 'versions', ...)` to get bluebird to work.

Strangely, this "bug" only became a problem after using bluebird in the project for a month or so. I haven't hunted down the culprit but it is are populating `global.process` with an `[object process]` 

I'm not sure if this is really the correct change, since this logic effects whether setImmediate/process.nextTick is used, and we probably want to use setImmediate in a React Native environment.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -385,8 +385,12 @@ var ret = {
     domainBind: domainBind
 };
 ret.isRecentNode = ret.isNode && (function() {
-    var version = process.versions.node.split(".").map(Number);
-    return (version[0] === 0 && version[1] > 10) || (version[0] > 0);
+    if (process.versions) {
+        var version = process.versions.node.split(".").map(Number);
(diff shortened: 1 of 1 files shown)


Ginden:
I would suggest changing `isNode`, not `isRecentNode`.

mlostekk:
This is blocking debugging with VSCode. @L8D how was your workaround looking like?

mlostekk:
I have fixed it temporarily like this

1. Install node modules
2. Open `bluebird.js`
3. Modify the code for `ret.isRecentNode` to

```
ret.isRecentNode = ret.isNode && (function() {
    var version = process.version.split(".").map(Number);
    return (version[0] === 0 && version[1] > 10) || (version[0] > 0);
})();
```

Issue #1529, opened 2018-05-30
OutOfMemory due to results in chain retained


peterlundberg:
1) What version of bluebird is the issue happening on?
3.5.1
2) What platform and version? (For example Node.js 0.12 or Google Chrome 32)
Node v8.11.1 on osx Darwin Kernel Version 17.5.0 x86_64

3) Did this issue happen with earlier version of bluebird?
Not sure

----

Using promise chains can cause unexpected OutOfMemory situations as the partial results in the chain seam to be retained until the entire chain is settled. 

For example in this `promise-chain-retain.js`:
```javascript
// global.Promise = require('bluebird');
console.log('Using', Promise.version || 'vanilla Promises');
logMemoryUse('before all');

Promise.resolve()
.then(() => {
  logMemoryUse('build result');
  const rows = [];
  for (let i = 0; i < 1000000; i += 1) {
    rows.push([`Example ${i}`, i, i * 2]);
  }
  return Promise.resolve({ rows });
})
.then((result) => {
  logMemoryUse('use result');
  // this then gets the large result
  return { len: result.rows.length };
})
.then((summaryResult) => {
  logMemoryUse('use summary');
  // this then should not need the 'result' and be free to reuse heap
  console.log(summaryResult);
})
.catch((err) => {
  console.log('caught error: ', err);
});

setTimeout(() => logMemoryUse('after all'), 3000);

function logMemoryUse(prefix = '') {
  global.gc();
  global.gc();
  const toMiB = b => (b / 1048576).toFixed(2);
  const mem = process.memoryUsage();
  console.log([
    'memoryUsage:',
    prefix.padEnd(20),
    `rss ${toMiB(mem.rss)} Mb`.padStart(20),
    // `heapTotal ${toMiB(mem.heapTotal)} Mb`.padStart(20),
    `heapUsed ${toMiB(mem.heapUsed)} Mb`.padStart(20),
  ].join('  '));
}
```

Using node 8 vanilla promises this works as expected in a chain and in "use summary" step the big result is no longer retained and can be GC:ed:

```
❯ NODE_ENV="production" node --expose-gc promise-chain-retain.js
Using vanilla Promises
memoryUsage:  before all                    rss 21.21 Mb    heapUsed 3.95 Mb
memoryUsage:  build result                  rss 21.28 Mb    heapUsed 3.94 Mb
memoryUsage:  use result                   rss 192.95 Mb    heapUsed 151.17 Mb
[...]

PR #1544, opened 2018-08-30
Fix memory being retained until promise queue is completely empty

Shingyx:
Achieved by draining each queue element in a separate function.

Resolves https://github.com/petkaantonov/bluebird/issues/1529. The issue provides a way to easily reproduce this issue.

I'm currently looking into adding a test to assert this behaviour.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
 git:
   depth: 5
 env:
-  - "NODE_FLAGS='' SCRIPT_FLAGS=''"
+  - "NODE_FLAGS='--expose-gc' SCRIPT_FLAGS=''"
 before_script:
 - git submodule update --init --recursive
 script: "node $NODE_FLAGS tools/test.js $SCRIPT_FLAGS"
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   ],
   "scripts": {
     "lint": "node scripts/jshint.js",
-    "test": "node tools/test.js",
+    "test": "node --expose-gc tools/test.js",
     "istanbul": "istanbul",
     "prepublish": "npm run generate-browser-core && npm run generate-browser-full",
     "generate-browser-full": "node tools/build.js --no-clean --no-debug --release --browser --minify",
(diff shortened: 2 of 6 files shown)


Shingyx:
I've now added unit tests, based on the code snippet from the linked issue. It exposed a new issue when `longStackTraces` is enabled, where somehow the `_trace` is being retained after the promise is fulfilled and finishes scheduling the next promise in the queue.

I fixed this by dereferencing `_trace` _after_ the promise is scheduled, to try minimise risk. Placing the dereference outside of the `if (BIT_FIELD_READ(LENGTH_MASK) > 0)` resulted in failing `generator.js` tests for some reason when running `npm test`, while the same tests passed when running `node tools/test --run=generator.js`...

benjamingr:
This generally LGTM but I won't have time to look at it until I'm back from holiday. Good find!

cc @petkaantonov @spion

package.json
```diff
@@ -19,7 +19,7 @@
   ],
   "scripts": {
     "lint": "node scripts/jshint.js",
-    "test": "node tools/test.js",
+    "test": "node --expose-gc tools/test.js",
```
Shingyx: To make Browserify happy, another option is to replace this commit with a `complete.exclude("v8");` inside browser_test_generator.js, if that's preferred.

.travis.yml
```diff
@@ -14,7 +14,7 @@ matrix:
 git:
   depth: 5
 env:
-  - "NODE_FLAGS='' SCRIPT_FLAGS=''"
+  - "NODE_FLAGS='--expose-gc' SCRIPT_FLAGS=''"
```
Shingyx: To make Browserify happy, another option is to replace this commit with a `complete.exclude("v8");` inside browser_test_generator.js, if that's preferred.

petkaantonov:
Why would equal code only when called through function not cause memory leak?

Shingyx:
When I think about it, it's not technically a memory leak. It's more that the `_drainQueue` function didn't free the memory it allocated until the queue is completely drained. I'll update the title. I can update the commit messages too if you like.

I think the issue occurs because of how the garbage collector works, at least in V8. I _think_ if memory is allocated inside the function `_drainQueue`, that memory will only be freed when `_drainQueue` completely finishes, i.e. when the queue is completely drained. By moving the memory allocation (i.e. `queue.shift()`) to a separate function `_drainQueueStep`, then memory can be freed more often.

PR #1548, opened 2018-09-07
Update dependencies with vulnerabilities

kibertoad:
Updated dependencies based on "npm audit" output.


package.json
```diff
@@ -44,16 +44,17 @@
     "baconjs": "^0.7.43",
     "bluebird": "^2.9.2",
```
kibertoad: Is this supposed to be old bluebird? It breaks down with the new one, though.

benjamingr:
I'm fine with the update but none of that is security related since all of it is dev dependencies. Bluebird doesn't bring any dependency itself as far as I recall

PR #1555, opened 2018-10-04
Update promise.each.md to fix formatting and spelling

rgeraldporter:
Fix for formatting and spelling. One example on the webpage is quite unreadable as the formatting is broken, this adjustment should make it readable again.


benjamingr:
Thanks a lot. Appreciated.

PR #1561, opened 2018-10-30
Fix embedded code snippet

bennyn:
There need to be line breaks before the embedded code snippet for "Alternatively in ES6". 

If there are no line breaks, then the programming language (`js`) used for the syntax highlighting will be rendered as part of the sentence in the docs:

**Bug**

image

You can see the problem here: http://bluebirdjs.com/docs/getting-started.html


benjamingr:
Thanks, will be out at next docs rebuild

PR #1571, opened 2018-12-18
Update jsdelivr url in docs

gricey432:
Currently this is 404ing on the live docs

Updated to be the same as suggested on the [install](http://bluebirdjs.com/docs/install.html) page


benjamingr:
Thanks!

PR #1584, opened 2019-02-24
.md syntax fix for coming-from-other-languages.md

elyobo:
Titles were lacking spaces, breaking linking from the ToC and making it hard to read.


benjamingr:
Thanks

PR #1594, opened 2019-05-04
Fix ES6 import strategy

Richienb:
Using the old import strategy, the following error is produced:

```js
Uncaught TypeError: Promise is not a constructor
```

It's been fixed in this PR.

### DIFF
--- a/docs/docs/getting-started.md
+++ b/docs/docs/getting-started.md
@@ -20,6 +20,12 @@ var Promise = require("bluebird");
 Alternatively in ES6 
 
 ```js
+import * as Promise from "bluebird";
+```
+
+If that ES6 import [doesn't work](https://github.com/petkaantonov/bluebird/pull/1594)
+
+```js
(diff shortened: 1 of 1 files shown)


benjamingr:
There's something wrong with your build - the way you fixed it too shouldn't technically work because the export is Promise and there is no named export (which is not destructuring)

Richienb:
@benjamingr That's strange. Anyways, here's a [link to the repo](https://github.com/ros-av/gui) that had this problem.

Gp2mv3:
We had the exact same issue in a React Native app

Richienb:
@benjamingr It seems as if the import can spontaneously stop and start working. As a result, we can just add the replacement to the docs instead of replacing the current code.

isaac-deephealth:
Coming from Node 22 with ESM and Typescript (NodeNext Module resolution). For some reason the default import must be named the same as the Exported member 

```import Bluebird from 'bluebird';```

PR #1596, opened 2019-05-13
unwrap ternary for strict mode compatibility

BryanCrotaz:
When using bluebird as a dependency and building with rollup, `rollup-plugin-commonjs` requires the dependency to work in strict mode.

This PR unrolls a nested ternary which is documented in many places as a code smell and isn't supported in strict mode.

### DIFF
--- a/src/debuggability.js
+++ b/src/debuggability.js
@@ -105,18 +105,24 @@ Promise.prototype._warn = function(message, shouldUseOwnTrace, promise) {
 
 Promise.onPossiblyUnhandledRejection = function (fn) {
     var domain = getDomain();
-    possiblyUnhandledRejection =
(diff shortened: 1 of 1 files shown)


petkaantonov:
What do you mean by strict mode?

marleen89:
Ffg

BryanCrotaz:
> What do you mean by strict mode?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

Ginden:
Ternaries are supported in strict mode, nested ternaries are supported too. This is likely an issue with your build system.

```
(function(){
'use strict';
return Math.random() > 0.5 ? Math.random() > 0.5 ? 1 : 2 : 3;
})()
```
Works.

PR #1605, opened 2019-07-16
jQuery did get it right, but it took five years.

amansheim-bm-net:


benjamingr:
Yeah jQuery 3 does the right thing here - thanks for this. I'll let Petka merge this since it requires a doc rebuild anyway

petkaantonov:
Maybe it could be rewritten a bit less cringe inducing

PR #1610, opened 2019-09-26
Add async/await example

Piccirello:


petkaantonov:
Even the mapping function could be async function but lgtm

PR #1627, opened 2019-11-27
Add fallback for unhandled rejection timer on Firefox.

### LINKED ISSUE #1623: Unhandled rejection check causes errors in Firefox
* Bluebird 3.7.0
* Firefox 71 (Nightly)

I recently upgraded from Bluebird 3.5.5 to 3.7.0, which happens to pull in the unhandled rejection handling from https://github.com/petkaantonov/bluebird/commit/60ef7a0e23fd320a11281f67c64a39ff95612ce9 (first added in 3.6.0).

The path which adds an `iframe`, grabs `setTimeout` from it, then removes it from the document does not appear to work in Firefox 71. Instead, calling this `setTimeout` fails with an `NS_ERROR_NOT_INITIALIZED` exception.

(As an aside, I am curious why you want to pull `setTimeout` from a throwaway frame like this... Why not just use the main document?)

For now, I will revert back to 3.5.5.

binki:
The default unhandled rejection timer is a setTimeout() from a removed
iframe. On Firefox, this throws NS_ERROR_NOT_INITIALIZED. This code
adds a test to verify that the setTimeout gleaned from the iframe is
functional and falls back to the global function if it isn’t.

This fixes support for rejecting promises on Firefox.

Fixes #1623.

### DIFF
--- a/src/debuggability.js
+++ b/src/debuggability.js
@@ -46,8 +46,13 @@ var deferUnhandledRejectionCheck;
         promises.length = 0;
     }
 
+    function fallbackDeferUnhandledRejectionCheck(promise) {
+        promises.push(promise);
+        setTimeout(unhandledRejectionCheck, 1);
+    }
+
     if (typeof document === "object" && document.createElement) {
-        deferUnhandledRejectionCheck = (function() {
+        (function() {
             var iframeSetTimeout;
 
             function checkIframe() {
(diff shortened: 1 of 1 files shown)


binki:
Playground at https://jsfiddle.net/m86xou54/ , see behavior in e.g. Firefox-71.0b12 which reproduces the issue.

Issue #1626, opened 2019-11-19
map() with concurrency ignores elements order


ArtemInvisMaliuha:
(This issue tracker is only for bug reports or feature requests, if this is neither, please choose appropriate channel from http://bluebirdjs.com/docs/support.html)

Please answer the questions the best you can:

1) What version of bluebird is the issue happening on?
3.7.0

2) What platform and version? (For example Node.js 0.12 or Google Chrome 32)
Node 10.16.3

3) Did this issue happen with earlier version of bluebird?
I dont know, but looks like yes ))

map() with concurrency option ignores elements order. It's unexpected behavior. mapSeries doesn't solving it because it doesn't have concurrency. I can make code donation.

benjamingr:
Thanks for opening this issue. This behavior s by design and has been discussed in the past. It is also explained in the docs.

ZaLiTHkA:
@benjamingr, I've been reading through historic issues related to this same topic, but I can't seem to find any clear indication as to how/why this is "by design". every time this comes up, the collaborator's generic response is "because that's how it is".

furthermore, [the documentation](http://bluebirdjs.com/docs/api/promise.map.html) *doesn't* do much to explain how the mapped array will be process, stating in a very generic way that...

>The order map calls the mapper function on the array elements is not specified, there is no guarantee on the order in which it'll execute the maper on the elements.

meanwhile, `MappingPromiseArray.prototype._drainQueue` itself inside `src/map.js` shows quite clearly what will happen..

1. start processing promises from the array start (as any "map" methods do in the JS world), until the defined concurrency limit is reached
2. pick the next promise from the array end as soon as one of the in flight promises settles

even looking at the tests provided in `test/mocha/map.js` it's clear to see that this is/was "known" to the Bluebird devs... this behaviour doesn't appear in the initial concurrency tests because these are run on an array with a length of 3 and `{ concurrency: 2 }`, which means it handles the first two, then the final.

going further down the same test file, there's a much more comprehensive concurrency test that mas an input of `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]` with `{ concurrency: 2 }` to an output of `[0, 1, 2, 3, 4, 10, 9, 8, 7, 6, 5]`.

the point here that really perplexes me is that this whole story could be fixed by changing a single line of code, simply making `MappingPromiseArray.prototype._drainQueue` use `queue.shift()` in place of `queue.pop()`.

[...]

PR #1631, opened 2019-12-17
Map concurrency retain array order

ZaLiTHkA:
The core of this PR simply changes the `_drainQueue` method used by the `map` method called when it's run with a concurrency limit, which fetches the next Promise in the array using `Array.shift()` instead of `Array.pop()`.

The point of this is to retain the standard behaviour of `map` and process each entry in the source array in the same order in which it was provided, regardless of whether or not a concurrent processing limit has been set, providing more consistent and predictable behaviour.

---

To ensure the unit tests cover this properly, I added an extra item to all typical test inputs, effectively changing `[1, 2, 3]` to `[1, 2, 3, 4]` and updated the order of the outputs for the final "big" test to match the order of the provided input.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -135,7 +135,7 @@ MappingPromiseArray.prototype._drainQueue = function () {
     var values = this._values;
     while (queue.length > 0 && this._inFlight < limit) {
         if (this._isResolved()) return;
-        var index = queue.pop();
+        var index = queue.shift();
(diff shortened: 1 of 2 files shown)


ZaLiTHkA:
closing this PR in favour of a better solution proposed in [this comment](https://github.com/petkaantonov/bluebird/issues/1626#issuecomment-596755882).

Issue #1635, opened 2020-01-21
toFastProperties is not working


GoBrianGo:
(This issue tracker is only for bug reports or feature requests, if this is neither, please choose appropriate channel from http://bluebirdjs.com/docs/support.html)

Please answer the questions the best you can:

1) What version of bluebird is the issue happening on?

2) What platform and version? (For example Node.js 0.12 or Google Chrome 32)

3) Did this issue happen with earlier version of bluebird?

(Write description of your issue here, stack traces from errors and code that reproduces the issue are helpful)

I found the function 'toFastProperties' is not working in Chromium/81.0.4034.0.

![image](https://user-images.githubusercontent.com/9709359/72773858-7ca6c100-3c43-11ea-86a6-66fa5dbe5a6f.png)

benjamingr:
https://github.com/sindresorhus/to-fast-properties/pull/13

@petkaantonov want me to update the bb impl as well?

petkaantonov:
sure

Ayushi96:
Is this issue closed in some released version ?

schultr81-ctrl:
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

schultr81-ctrl:
Get Outlook for Android<https://aka.ms/AAb9ysg>

schultr81-ctrl:
Check out the stunning view of the sky in 42° 38'N 92° 27'W (Cedar Falls, IA, United States) at May 25, 2026 4:14 AM, captured using the Sky Tonight Image Editor. #MySkyTonight #SkyTonight #GoStargazing #StarWalk https://get.skytonight.app/share

Get Outlook for Android<https://aka.ms/AAb9ysg>

schultr81-ctrl:
Keeping up with the latest astronomy news in the stargazing app #StarWalk. #SkyTonight #GoStargazing #StarWalk https://starwalk.space/th/en/bright-mars-2025

Get Outlook for Android<https://aka.ms/AAb9ysg>

schultr81-ctrl:
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

schultr81-ctrl:
[...]

PR #1636, opened 2020-01-23
Fix toFastProperties

benjamingr:


vitaly-t:
Does this fix #1635? Just so, there was no reference provided here, and that issue wasn't updated either.

benjamingr:
Yes

iahu:
hi all, still find `eval` keyword on `toFastProperties`, it seems you are not public a new version since year 2019.

please public this PR to NPM, to fix this warning, thanks.

```bash
bundles ./src/**/*.ts, !./src/**/*.test.ts, !./src/renderer, !./src/data → dist/esm...
(!) Use of eval is strongly discouraged
https://rollupjs.org/guide/en/#avoiding-eval
../../node_modules/bluebird/js/release/util.js
203:     ic();
204:     return obj;
205:     eval(obj);
         ^
206: }
created dist/esm in 2.7s
```

PR #1648, opened 2020-06-12
docs: Fix simple typo, standart -> standard

timgates42:
There is a small typo in test/browser/es5-shim.js.

Should read `standard` rather than `standart`.


benjamingr:
Thanks!

PR #1651, opened 2020-07-10
Fix benchmarks since iterate memory pages is expensive

Aschen:
I got some strange results by benchmarking promises with those benchmarks. Performances seemed to be worst after Node.js 10.

I reported the issue on Node and one of the developers finds out: https://github.com/nodejs/node/issues/33384

It's because default page size has been divided by 2 after Node.js 10 so iterate over memory pages with `process.memoryUsage()` to get the `rss` is at least 2 times slower.  

I fixed the benchmark so they don't take the pages iteration time into account.

I also updated the benchmark results.


Aschen:
Should we merge this one so subsequent benchmarks will be correct?

benjamingr:
@petkaantonov ?

Issue #1714, opened 2024-08-09
A11y : focusable element - html lang attribute - link underline


mariamesylla:
(This issue tracker is only for bug reports or feature requests, if this is neither, please choose appropriate channel from http://bluebirdjs.com/docs/support.html)

Please answer the questions the best you can:

1) What version of bluebird is the issue happening on? "version": "3.7.2"

2) What platform and version? Google Chrome Version 123.0.6312.87 (Official Build)

3) Did this issue happen with earlier version of bluebird? Yes

(Write description of your issue here, stack traces from errors and code that reproduces the issue are helpful)

<html> element must have a lang attribute

Add a lang attribute to the html element (e.g. <html lang="en">) whose value represents the primary language of document.

Make sure you identify a language in the opening <html> element and spell the attribute correctly. For example, if the primary language of a document is English, you could specify the language as follows:

mariamesylla:
Assigned issue to @mariamesylla

PR #1715, opened 2024-08-12
Ally/fix

mariamesylla:
**Issue added:** https://github.com/petkaantonov/bluebird/issues/1714

Ensure aria-hidden elements are not focusable nor contain focusable elements
[more informationLink opens in a new window](https://dequeuniversity.com/rules/axe/4.9/link-in-text-block?application=AxeChrome)
Screenshot 2024-08-09 at 1 27 27 PM

Ensures every HTML document has a lang attribute in https://bluebirdjs.com/docs/getting-started.html

Screenshot 2024-08-09 at 12 50 11 PM

Ensure links are distinguished from surrounding text in a way that does not rely on color
Screenshot 2024-08-09 at 1 27 43 PM


mariamesylla:
@petkaantonov

PR #1732, opened 2026-09-20
Copy custom properties onto Promise.method wrappers

### LINKED ISSUE #1484: Promise.method does not retain custom properties

3.5.1

*

*

(Write description of your issue here, stack traces from errors and code that reproduces the issue are helpful)

`Promise.method` does not retain custom properties, for example

```js
const Promise = require("bluebird")
const fn = () => {}
fn.foo = "bar"
Promise.promisify(fn).foo
// expected: bar, returns: bar
Promise.method(fn).foo
// expected: bar, returns: undefined
```

On first glance we can use `util.copyDescriptors` to retain custom properties as it is done in `promisify`, however I am wondering about the justification of not preserving custom properties in `Promise.method`. And if it is not a bug, how to implement it on userland.

dyk1454683243-sudo:
Fixes #1484

## Cause

`Promise.method` wraps the original function in a new anonymous function and returns that wrapper as-is. Own properties on the input (for example `fn.foo = "bar"`) are therefore dropped.

`Promise.promisify` already copies those properties onto its wrapper with `util.copyDescriptors` and a filter that skips built-in function keys (`length`, `name`, `prototype`, and the rest of the same skip list). `Promise.method` never did the same thing.

## Fix

After creating the `Promise.method` wrapper, copy own property descriptors from the original function with `util.copyDescriptors` and the same built-in skip list used by `promisify`.

```js
const fn = () => {}
fn.foo = "bar"
Promise.promisify(fn).foo // "bar"
Promise.method(fn).foo    // was undefined; now "bar"
```

## Tests

Added a regression test in `test/mocha/method.js` for the issue repro: a function with `fn.foo = "bar"` keeps `.foo` after `Promise.method(fn)`, and the wrapped function still fulfills correctly.

`test/mocha/method.js`: 11 passing, including the new case. Repro now prints `"bar"` for both `Promise.promisify(fn).foo` and `Promise.method(fn).foo`.

Full `node --expose-gc tools/test.js` on Node 22 reports `generator.js` (`Long stack traces with coroutines as context` expects 2 frames and sees 9). The same assertion fails on current `master`. Isolated `promisify.js` also has a pre-existing arity failure on Node 22 (`0 == 3`). Neither is caused by this change.


dyk1454683243-sudo:

PR #1733, opened 2026-09-20
Treat .done() chains as handled for runaway warnings

### LINKED ISSUE #1491: .done() signifies a promise is handled; chains that end in it are not runaway promises
3.5.1

NodeJS 8.9.1

Unknown - new usage.

Currently, there seems to be no way for a library that wishes to use promises but expose a callback api without causing floods of false-positive runaway promise warnings if a consuming library then wraps those callbacks in other bluebird promises.

It seems the semantics of .done() indicate that the promise is absolutely at an end and fully handled, so we have been using it for promises that perform setup work inside things like streams, which communicate their readiness over EventEmitter events rather than exporting a promise.  As such, any promise chain in which all branches end in a .done() should not be considered a runaway promise error.

After a few minutes of experimentation, I was able to make a trial implementation of this in debuggability.js by calling this function from checkForgottenReturns against ```promiseCreated```
```
function checkEventuallyFinal(promise) {
    if (promise._isFinal()) return true;
    if (promise._promise0 && !checkEventuallyFinal(promise._promise0)) return false;
    var len = (promise._bitField & 65535);
    for(var i = 1; i < len; i++) {
        var child = promise._promiseAt(i);
        if (!checkEventuallyFinal(child)) return false;
    }
    return true;
}
```

I have tested this out against our existing code, and it reliably silences the error when the promises are fully handled and end in .done(), and still warn when it detects 'true' runaway promises.

If performance is an issue, a bit field could be added indicating if all children are final, and propagated on each .done() call.

dyk1454683243-sudo:
Fixes #1491

## Cause

`.done()` marks its continuation `_isFinal()`, meaning the chain is finished and handled. Forgotten-return / runaway detection does not look at that.

`checkForgottenReturns` records only the **first** promise created inside a handler. A typical library-that-uses-promises-but-exposes-callbacks pattern is:

```js
function close() {
    connectPromise.then(doCloseWork).done(emitClose, emitError);
}

outer.then(function() {
    api.close(); // creates a pending chain that ends in .done()
}).then(function() { /* consumer chained this */ });
```

The first created promise is still pending and has children, so it is not `_isFinal()`. Because the consumer chained the outer handler, Bluebird emits “a promise was created in a handler but was not returned from it” even though every branch ends in `.done()`.

`.asCallback()` is a different contract: it attaches a nodeback and returns the original promise, and does not set `_isFinal()`. This change only treats `.done()` as a terminator.

Unhandled rejection behavior is unchanged. A rejection that reaches `.done()` without an errback is still a fatal error, not an `unhandledRejection`.

## Fix

In `checkForgottenReturns`, walk `promiseCreated` and suppress the warning when every branch is eventually `_isFinal()` (including following promises). The walk only runs on the existing debug warning path.

## Tests

`test/mocha/github-1491.js` covers:

- a pending chain ending in `.done()` is not reported as runaway
- a true forgotten return still warns
- all branches `.done()` → no warning; mixed branches → warning
- the EventEmitter / callback-API wrapping pattern from the issue
- rejected `.done()` is fatal and is not reported as unhandled

`node --expose-gc tools/test.js --run=github-1491`: 6 passing
`node --expose-gc tools/test.js --run=done`: 5 passing
`node --expose-gc tools/test.js --run=unhandled_rejections`: 56 passing

## Suite notes

Full `node --expose-gc tools/test.js` on Node 22 fails only:

- `generator.js` — `Long stack traces with coroutines as context` expects 2 frames and sees 9
- `promisify.js` — arity `should be original - 1` (`0 == 3`)

The same two assertions fail on current `master` with this change reverted. They are pre-existing on Node 22, not caused by this PR. All other mocha files pass, including `github-1491.js`.


dyk1454683243-sudo:

PR #1734, opened 2026-09-20
Fix Promise.each calling .then twice on thenables

### LINKED ISSUE #1455: .then called twice on promises passed to Promise.each
(Bluebird 3.5.0, node 8.4.0)

Whether you pass it a promise of an array or an array of promises, Promise.each will `.then` all of those promises twice. Here's a small code sample that demonstrates this:
```js
var Promise = require("bluebird");

function makeFakePromise(val) {
  var p = {
    count: 0,
    then: function(f) {
      p.count += 1;
      return Promise.resolve(val).then(f);
    },
  };
  return p;
}

var arrayOfPromises = [0, 1, 2].map(makeFakePromise);
var promisedArray = makeFakePromise(arrayOfPromises);

Promise.each(promisedArray, function() {}).then(function() {
  console.log(promisedArray.count); // 2
  console.log(arrayOfPromises[0].count); // 2
  console.log(arrayOfPromises[1].count); // 2
  console.log(arrayOfPromises[2].count); // 2
});
```
Personally I want to believe this is something that Bluebird should be allowed to do safely, and I would understand closing this as no action. However, libraries like [knex](https://github.com/tgriesser/knex/pull/2226) give a `.then` call special treatment, in this case running a database query for each call. As the link shows, knex has decided not to fix it on their end, at least for now. There are workarounds, and one could argue that knex's query building interface is not a proper promise and should not be given to bluebird for handling, but this is very easy to get wrong in practice.

The above code sample was extracted from a larger test file that counted `.then` calls for every Bluebird method being called in various ways. I believe Promise.each and the equivalent prototype method are the only parts of Bluebird that have this behavior.

dyk1454683243-sudo:
Fixes #1455

## Cause

`Promise.each` / `.each` run the iterator through `Promise.reduce`, which assimilates each input thenable once via `tryConvertToPromise`. After reduce finishes they called `Promise.all` on the original input to build the result array of resolved values. `Promise.all` assimilates those same thenables again.

That second `.then` is what the #1455 repro measures. Libraries such as knex treat each `.then` as a new side effect (a new query). Bluebird promises were mostly unaffected because they go through `_then` rather than the public `.then`.

`Promise.mapSeries` already returns values collected during reduce and did not have this extra `Promise.all`.

## Fix

Collect each resolved input value during the reduce pass (`_preservedValues`) and resolve to that array. Drop the trailing `Promise.all`.

- `Promise.each` still resolves to the original elements' values, not the iterator outputs.
- Empty input still resolves to `[]`.
- Each input thenable (and a thenable-of-array) is assimilated once.

## Tests

`test/mocha/github-1455.js` covers:

- array of thenables (issue repro)
- thenable of an array of thenables (issue repro)
- `Promise.prototype.each`
- iterator still receives resolved values; result ignores mapper output
- empty thenable-of-array is only then'd once
- rejected thenable is only then'd once

On current master the first assertion fails with `2 !== 1`.

```
node --expose-gc tools/test.js --run=github-1455   # 6 passing
node --expose-gc tools/test.js --run=each          # 13 passing
node --expose-gc tools/test.js --run=reduce        # 294 passing
node --expose-gc tools/test.js --run=collections_thenables  # 40 passing
node --expose-gc tools/test.js --run=map           # 20 passing
node --expose-gc tools/test.js --run=generator     # 36 passing
```

## Suite notes

Full `node --expose-gc tools/test.js` on Node 22 fails only `promisify.js` (`arity should be original - 1`, `0 == 3`). The same assertion fails on current `master` with this change reverted. It is pre-existing on Node 22, not caused by this PR.

Fork: https://github.com/dyk1454683243-sudo/bluebird/pull/3


dyk1454683243-sudo:

PR #1735, opened 2026-09-20
Fix coroutine sync-throw stack traces when longStackTraces is on

### LINKED ISSUE #1488: Incorrect stack trace when coroutine throws sync and long stack traces enabled

v3.5.1

Node 6.12.2, Mac OS 10.12.6

Unknown

---

This is a very small nit.

When long stack traces are enabled, the stack trace for an error thrown synchronously in a coroutine includes "From previous event" when in fact everything happened synchronously.

```js
const Promise = require('bluebird');
Promise.longStackTraces();

Promise.coroutine(function *() {
	throw new Error('foo');
})();
```

Produces error:

```
Error: foo
    at /test/test.js:5:8
    at next (native)
From previous event:
    at Object.<anonymous> (/test/test.js:6:3)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:383:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:496:3
```

Without long stack traces enabled, error trace is:

```
Error: foo
    at /test/test.js:5:8
    at next (native)
    at tryCatcher (/test/node_modules/bluebird/js/release/util.js:16:23)
    at PromiseSpawn._promiseFulfilled (/test/node_modules/bluebird/js/release/generators.js:97:49)
    at /test/node_modules/bluebird/js/release/generators.js:201:15
    at Object.<anonymous> (/test/test.js:6:3)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:383:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:496:3
```

The source of the problem appears to be [this line](https://github.com/petkaantonov/bluebird/blob/master/src/generators.js#L140) where `._rejectCallback()` is always called with `synchronous = false` regardless of whether error is thrown sync or not.

dyk1454683243-sudo:
Fixes #1488

## Cause

`Promise.coroutine` / `PromiseSpawn._continue` always rejected with `_rejectCallback(error, false)`.

The second argument is `synchronous`. When it is `false`, long stack traces attach the spawned promise's captured stack as a new hop. That produces a `From previous event` section even when the generator threw on its first turn and nothing was yielded.

```js
Promise.longStackTraces();
Promise.coroutine(function *() {
    throw new Error('foo');
})();
```

## Fix

Track whether the spawn is still on its first synchronous turn (`_synchronous`). Pass that flag to `_rejectCallback`. After a real yield (pending proxy or `_async.invoke`), the flag is cleared so later async throws still get the extra hop.

## Tests

`test/mocha/github-1488.js` covers:

- sync throw has no `From previous event`
- throw site is still present
- throw after `yield Promise.delay` still has `From previous event`
- coroutine started from a `.then` handler still keeps caller context

The existing 4-level coroutine long-stack fixture throws in the innermost generator without yielding. After this fix that caller stays on the same hop, so `assertLongTrace` is now `3+1, [4, 2, 2]` instead of `4+1, [2, 2, 2, 2]`.

```
node --expose-gc tools/test.js --run=github-1488       # 4 passing
node --expose-gc tools/test.js --run=generator         # 36 passing
node --expose-gc tools/test.js --run=long_stack_traces # 37 passing
```

## Suite notes

Full `node --expose-gc tools/test.js` on Node 22 fails only `promisify.js` (`arity should be original - 1`, `0 == 3`). The same assertion fails on current `master`. It is pre-existing on Node 22, not caused by this PR.

Fork PR: https://github.com/dyk1454683243-sudo/bluebird/pull/4
Fork compare: https://github.com/dyk1454683243-sudo/bluebird/compare/master...cursor/fix-coroutine-sync-throw-stack-1488-1435


dyk1454683243-sudo:

